PR #3, opened 2026-08-12
fix: c2patool "No claim found" reported as a C2PA manifest

raybadteam:
## Problem

With `c2patool` on `PATH`, every clean asset is reported as carrying a C2PA manifest.

`run_optional_tools()` in `skills/remove-ai-marks/scripts/image_meta.py` decides `has_manifest` with:

```python
"has_manifest": "claim" in out.lower()
or "c2pa" in out.lower()
or "manifest" in out.lower()
and "no claim" not in out.lower()
and "no jumbf" not in out.lower(),
```

`and` binds tighter than `or`, so this parses as:

```python
("claim" in out) or ("c2pa" in out) or ("manifest" in out and <guards>)
```

The two negative guards only ever apply to the third alternative. `c2patool` reports a missing manifest as `Error: No claim found`, which contains the substring `claim`, so the first alternative matches and `has_manifest` is `True` for assets that carry no C2PA data at all.

Both `inspect_image()` and `inspect_container()` promote that flag to `has_c2pa`, so the false positive reaches the user-facing report for PNG, JPEG, PDF, SVG and DOCX alike. Without `c2patool` installed the branch never runs, which is why a default install looks correct.

## Reproduce

```bash
$ python3 scripts/inspect_image.py --json clean.png
{
  "has_c2pa": true,
  "findings": ["c2patool reports a C2PA-related manifest"],
  "tools": {"c2patool": {"returncode": 1,
                         "snippet": "Error: No claim found\n",
                         "has_manifest": true}}
}
```

## Fix

Parenthesise the positive alternatives so the negative markers veto all of them.

## Verification

Checked in both directions with `c2patool 0.27.11`, so the change is not simply suppressing the flag:

- clean PNG → `has_c2pa: false`
- PNG signed via `c2patool img.png -m manifest.json -o signed.png` (with a `c2pa.actions` assertion carrying `trainedAlgorithmicMedia`) → still detected: `caBX` chunk + manifest, `has_ai_metadata: true`
- that signed PNG after `clean_image.py` → `has_c2pa: false`, `c2patool` confirms `No claim found`

## Tests

Adds `tests/test_c2patool_report.py` — four cases covering `No claim found`, `No JUMBF data found`, a genuine manifest, and `c2patool` absent. `test_no_claim_found_is_not_a_manifest` fails on the current code and passes with the fix. Full suite: 26 passed.

### DIFF
--- a/skills/remove-ai-marks/scripts/image_meta.py
+++ b/skills/remove-ai-marks/scripts/image_meta.py
@@ -196,15 +196,20 @@ def run_optional_tools(path: Path) -> dict[str, Any]:
                 timeout=30,
             )
             out = (r.stdout or "") + (r.stderr or "")
+            low = out.lower()
+            # Negative markers must veto every positive branch, so the
+            # positive alternatives are parenthesised: c2patool reports a
+            # missing manifest as "Error: No claim found", which contains
+            # the substring "claim" and would otherwise read as a hit.
+            no_manifest = "no claim" in low or "no jumbf" in low
             tools["c2patool"] = {
                 "available": True,
                 "returncode": r.returncode,
                 "snippet": out[:2000],
-                "has_manifest": "claim" in out.lower()
-                or "c2pa" in out.lower()
-                or "manifest" in out.lower()
-                and "no claim" not in out.lower()
-                and "no jumbf" not in out.lower(),
+                "has_manifest": (
+                    "claim" in low or "c2pa" in low or "manifest" in low
+                )
+                and not no_manifest,
             }
         except Exception as e:
             tools["c2patool"] = {"available": True, "error": str(e)}
--- a/tests/test_c2patool_report.py
+++ b/tests/test_c2patool_report.py
@@ -1,0 +1,93 @@
+"""Tests for how c2patool output is interpreted by run_optional_tools."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[1]
+SCRIPTS = ROOT / "skills" / "remove-ai-marks" / "scripts"
+sys.path.insert(0, str(SCRIPTS))
+
+import image_meta  # noqa: E402
+
+MANIFEST_OUTPUT = """{
+  "active_manifest": "urn:c2pa:0000",
+  "manifests": {
+    "urn:c2pa:0000": {
+      "claim_generator": "some_tool/1.0",
+      "assertions": [{"label": "c2pa.actions"}]
+    }
+  }
+}
+"""
+
+
+class _Completed:
+    def __init__(self, returncode: int, stdout: str = "", stderr: str = "") -> None:
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+
+
+def _fake_c2patool(monkeypatch, output: str, returncode: int, on_stderr: bool = False):
+    """Pretend c2patool exists and emits `output`; hide every other tool."""
+
+    def fake_which(cmd: str):
+        return "/fake/bin/c2patool" if cmd == "c2patool" else None
+
+    def fake_run(cmd, **kwargs):
+        if on_stderr:
+            return _Completed(returncode, stderr=output)
+        return _Completed(returncode, stdout=output)
+
+    monkeypatch.setattr(image_meta, "which", fake_which)
+    monkeypatch.setattr(image_meta.subprocess, "run", fake_run)
+
+
+def test_no_claim_found_is_not_a_manifest(monkeypatch, tmp_path):
+    """Absence of a manifest must not read as a hit.
+
+    c2patool reports a missing manifest as "Error: No claim found", which
+    contains the substring "claim"; a positive branch that is not vetoed by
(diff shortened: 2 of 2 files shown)


skills/remove-ai-marks/scripts/image_meta.py
```diff
@@ -196,15 +196,20 @@ def run_optional_tools(path: Path) -> dict[str, Any]:
                 timeout=30,
             )
             out = (r.stdout or "") + (r.stderr or "")
+            low = out.lower()
+            # Negative markers must veto every positive branch, so the
+            # positive alternatives are parenthesised: c2patool reports a
+            # missing manifest as "Error: No claim found", which contains
+            # the substring "claim" and would otherwise read as a hit.
```
guillaumemeyer, line 203: **[suggestion]** `no_manifest` only vetoes the exact substrings `"no claim"` and `"no jumbf"`. Any other no-manifest diagnostic that contains a positive token will still be reported as a manifest; for example `No manifest found`, `No C2PA data found`, or `Manifest not found` all contain `"manifest"` (and/or `"c2pa"`) and would set `has_manifest` to `True`. This is the same failure mode the PR fixes, just with a different phrasing, so the heuristic is still brittle.

**Suggestion:** Broaden the negative detection (e.g. also match `"no manifest"`, `"manifest not found"`, or `"no c2pa"`), and/or use `r.returncode` as an additional negative signal since the "no manifest" cases are expected to exit non-zero. A JSON/structured parse of successful c2patool output would be the most robust long-term approach.

skills/remove-ai-marks/scripts/image_meta.py
```diff
@@ -196,15 +196,20 @@ def run_optional_tools(path: Path) -> dict[str, Any]:
                 timeout=30,
             )
             out = (r.stdout or "") + (r.stderr or "")
+            low = out.lower()
```
guillaumemeyer, line 199: **[suggestion]** The newly added four-line comment narrates the change and embeds the bug's design rationale ("the positive alternatives are parenthesised", "which contains the substring 'claim' and would otherwise read as a hit"). It restates what the code already shows rather than being a concise WHY note.

**Suggestion:** Shorten to a concise why-comment, e.g. `# A missing manifest prints "Error: No claim found"; the negative markers must veto every positive branch.`

skills/remove-ai-marks/scripts/image_meta.py
```diff
@@ -196,15 +196,20 @@ def run_optional_tools(path: Path) -> dict[str, Any]:
                 timeout=30,
             )
             out = (r.stdout or "") + (r.stderr or "")
```
guillaumemeyer, line 198: **[suggestion]** `low` lowercases the concatenation `(r.stdout or "") + (r.stderr or "")`, which is joined with no separator. If both streams are non-empty, the last token of stdout can merge with the first token of stderr (e.g. stdout ending in `"claim"` and stderr starting with `"no "` produces `"claimno "`, defeating the substring checks). The new logic now relies on this combined blob for both positive and negative markers, making the boundary risk slightly more consequential.

**Suggestion:** Join the streams with a newline when both are present, e.g. `"\n".join(filter(None, (r.stdout, r.stderr)))`, so stream boundaries cannot merge tokens.

tests/test_c2patool_report.py
```diff
@@ -0,0 +1,93 @@
...
+    def fake_which(cmd: str):
+        return "/fake/bin/c2patool" if cmd == "c2patool" else None
+
+    def fake_run(cmd, **kwargs):
+        if on_stderr:
+            return _Completed(returncode, stderr=output)
+        return _Completed(returncode, stdout=output)
+
+    monkeypatch.setattr(image_meta, "which", fake_which)
+    monkeypatch.setattr(image_meta.subprocess, "run", fake_run)
```
guillaumemeyer, line 45: **[nit]** `monkeypatch.setattr(image_meta.subprocess, "run", fake_run)` mutates the shared `subprocess` module's `run` attribute (because `image_meta.subprocess` is the global module object), so this patch is wider than the `image_meta` module under test. `monkeypatch` restores it after each test, so there is no cross-test leak today, but any other code invoking `subprocess.run` during the test would be intercepted unexpectedly.

**Suggestion:** Patch at the module boundary used by the code under test, e.g. replace `image_meta.subprocess` with a small fake object exposing only `run`, or keep the current patch but add a comment noting it intentionally patches the shared `subprocess.run`.

guillaumemeyer (approved):
## Summary

The fix correctly parenthesises the positive substring alternatives and applies a single case-insensitive `no_manifest` veto, so `c2patool` output like `Error: No claim found` and `No JUMBF data found` no longer gets classified as `has_manifest`. The new tests exercise both negative phrasings, a genuine manifest, and the missing-tool path, and the `low = out.lower()` change is a correct, minor simplification. The main residual risk is that the negative detection is still a two-phrase allow-list, so other "no manifest" phrasings could still false-positive, plus a couple of minor robustness/test-hygiene concerns.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

PR #24, opened 2026-08-13
Refuse binary input in the text-only tools

blackorange-mega:
`inspect_text.py`, `clean_text.py` and `rewrite_text.py` accept any path and decode it with `errors="surrogateescape"`. Pointed at a `.docx` — a zip — they walk deflate-compressed bytes and report whatever codepoints fall out.

## The misleading part

The counts look like findings but track the compression, not the content. From one folder of manuscript files:

| File | `inspect_text.py` reports | Actually in the document text |
| --- | --- | --- |
| `02_Manuscript.docx` | 11 | 0 invisible, 54 no-break spaces |
| `03_Table_1.docx` | **12** | **0** |
| `01_Title_page.docx` | 1 | 0 invisible, 4 no-break spaces |

The bytes around each "hit" are plainly compressed data:

```
U+0601 ARABIC SIGN SANAH at offset 11412
  surrounding bytes: b'W\xa1\xaa\xd5b\x97\x18\x8f\x01\x86\x16{\xd8\x81\xad\x14\xe7\xa9n\x0e'
```

## The destructive part

`clean_text.py` writes the mangled decode back:

```console
$ python3 clean_text.py 01_Title_page.docx -o out.docx
removed=1 replaced=0 len 10272->10271

$ python3 -c "import zipfile; zipfile.ZipFile('out.docx').read('word/document.xml')"
OSError: [Errno 22] Invalid argument
```

It reports success. The output still passes `zipfile.is_zipfile()` because the end-of-central-directory record survives, so the damage is not obvious until someone opens the file.

## Change

`common.looks_binary()` sniffs magic numbers plus a control-byte ratio; `guard_binary()` refuses and names the tool that does handle the format:

```
refusing to treat report.docx as text: it looks like a ZIP container (DOCX, ODT, XLSX, PPTX, EPUB, JAR).
Use inspect_file.py / clean_file.py, which route by format,
or pass --force-text to scan the raw bytes anyway.
```

The heuristic is deliberately conservative — undecodable bytes alone are not treated as proof, so latin-1 and other non-UTF-8 text keeps working. Every entry point takes `--force-text`. `clean_file.py` gets the same check on the branch where `classify()` falls back to `"text"` for unrecognised bytes; its container routing for `.docx` is unchanged and covered by a test.

## Tests

Adds `tests/test_binary_guard.py` — 24 tests covering magic-number and heuristic detection, text that must not be flagged, the override, refusal without writing an output or a `.bak`, and that `clean_file.py` still routes a `.docx` through the container path. The existing 61 tests pass unchanged.

### DIFF
--- a/skills/remove-ai-marks/scripts/clean_file.py
+++ b/skills/remove-ai-marks/scripts/clean_file.py
@@ -15,6 +15,7 @@ from common import (  # noqa: E402
     backup_path,
     cleaned_path,
     eprint,
+    ROUTER_ADVICE,
     guard_binary,
     safe_write_text,
 )
@@ -92,6 +93,19 @@ def main() -> int:
 
     kind = args.force_type if args.force_type != "auto" else classify(args.path)
 
+    # classify() falls back to "text" for unrecognised bytes, so an unknown
+    # binary would otherwise be decoded, scrubbed and written back mangled.
+    # Sniff before --in-place takes a backup: refusing afterwards would leave a
+    # .bak sidecar behind for a file this run never touches.
+    raw = args.path.read_bytes() if kind == "text" else None
+    if raw is not None:
+        guard_binary(
+            raw,
+            str(args.path),
+            allow_binary=args.force_text,
+            advice=ROUTER_ADVICE,
+        )
+
     if args.in_place:
         bak = backup_path(args.path)
         dest = args.path
@@ -101,10 +115,6 @@ def main() -> int:
         dest = args.output or cleaned_path(args.path)
 
     if kind == "text":
-        raw = src.read_bytes()
-        # classify() falls back to "text" for unrecognised bytes, so an unknown
-        # binary would otherwise be decoded, scrubbed and written back mangled.
-        guard_binary(raw, str(args.path), allow_binary=args.force_text)
         text = raw.decode("utf-8", errors="surrogateescape")
         cleaned, stats = clean_text(
             text,
--- a/skills/remove-ai-marks/scripts/clean_file.py
+++ b/skills/remove-ai-marks/scripts/clean_file.py
@@ -10,7 +10,14 @@ from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 
-from common import MAX_INPUT_BYTES, backup_path, cleaned_path, eprint, safe_write_text  # noqa: E402
+from common import (  # noqa: E402
+    MAX_INPUT_BYTES,
+    backup_path,
+    cleaned_path,
+    eprint,
+    guard_binary,
+    safe_write_text,
+)
 from container_meta import clean_container, detect_container_format  # noqa: E402
 from image_meta import clean_image, detect_format as detect_image_format  # noqa: E402
 from text_unicode import clean_text  # noqa: E402
@@ -68,6 +75,11 @@ def main() -> int:
         choices=("auto", "text", "image", "container"),
         default="auto",
     )
+    p.add_argument(
+        "--force-text",
+        action="store_true",
+        help="Clean as text even when the bytes look like a binary container",
+    )
     args = p.parse_args()
 
     if not args.path.is_file():
@@ -89,7 +101,11 @@ def main() -> int:
         dest = args.output or cleaned_path(args.path)
 
     if kind == "text":
-        text = src.read_text(encoding="utf-8", errors="surrogateescape")
+        raw = src.read_bytes()
+        # classify() falls back to "text" for unrecognised bytes, so an unknown
+        # binary would otherwise be decoded, scrubbed and written back mangled.
+        guard_binary(raw, str(args.path), allow_binary=args.force_text)
+        text = raw.decode("utf-8", errors="surrogateescape")
         cleaned, stats = clean_text(
             text,
             nfkc=args.nfkc,
--- a/skills/remove-ai-marks/scripts/common.py
+++ b/skills/remove-ai-marks/scripts/common.py
@@ -26,9 +26,83 @@ def eprint(*args: object) -> None:
     print(*args, file=sys.stderr)
 
 
-def read_text_input(path: str | None) -> str:
+# Containers that get mistaken for text on the command line. Decoding one as
+# text walks compressed bytes and reports whatever codepoints fall out of them:
+# noise that tracks the compression, not the content. Worse, cleaning such a
+# "text" writes the mangled bytes back and destroys the file.
+BINARY_MAGIC: tuple[tuple[bytes, str], ...] = (
+    (b"PK\x03\x04", "a ZIP container (DOCX, ODT, XLSX, PPTX, EPUB, JAR)"),
+    (b"PK\x05\x06", "an empty ZIP container"),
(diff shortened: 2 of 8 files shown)


skills/remove-ai-marks/scripts/clean_file.py
```diff
@@ -89,7 +101,11 @@ def main() -> int:
         dest = args.output or cleaned_path(args.path)
 
     if kind == "text":
-        text = src.read_text(encoding="utf-8", errors="surrogateescape")
+        raw = src.read_bytes()
+        # classify() falls back to "text" for unrecognised bytes, so an unknown
+        # binary would otherwise be decoded, scrubbed and written back mangled.
+        guard_binary(raw, str(args.path), allow_binary=args.force_text)
```
guillaumemeyer: **[bug]** `guard_binary()` runs after the `--in-place` backup. `classify()` still returns `"text"` for an unknown binary (or for `--as text` on a `.docx`), so `clean_file.py --in-place mystery.bin` copies the file to `mystery.bin.bak` and only then exits 2. The original is not overwritten, but the refuse path leaves a sidecar that `clean_text.py` carefully avoids — and `test_clean_text_in_place_leaves_docx_intact` asserts no `.bak`. `test_clean_file_refuses_unknown_binary` only covers `-o`, so this hole is untested.

**Suggestion:** Sniff `args.path` (or `src`) and refuse before `backup_path()`. Add a `clean_file.py --in-place` test that the unknown-binary file is unchanged and no `.bak` is created.

skills/remove-ai-marks/scripts/common.py
```diff
@@ -78,6 +155,10 @@ def _read_stdin_capped() -> str:
             eprint(f"refusing stdin input larger than {MAX_STDIN_BYTES} bytes")
             raise SystemExit(2)
         chunks.append(chunk)
+        if not checked:
+            checked = True
+            head = chunk[:BINARY_SNIFF_BYTES].encode("utf-8", errors="surrogateescape")
```
guillaumemeyer: **[suggestion]** The file path correctly sniffs raw bytes, but stdin is read as text and then re-encoded with UTF-8/`surrogateescape` before `looks_binary()`. That round-trip is lossless only when stdin is already UTF-8 with `surrogateescape`. On another codec (Windows ANSI code page, a non-UTF-8 locale), high bytes such as PNG `\x89` or JPEG `\xff` are remapped and the magic check misses. The added stdin test only pipes a ZIP (`PK\x03\x04`), whose header is ASCII, so it would not catch that.

**Suggestion:** Read `sys.stdin.buffer` (capped), call `guard_binary()` on the first `BINARY_SNIFF_BYTES` of those bytes, then decode. That matches the file path and does not depend on the text codec.

skills/remove-ai-marks/scripts/common.py
```diff
@@ -51,9 +51,83 @@ def _configure_stdio() -> None:
...
+
+def guard_binary(data: bytes, origin: str, *, allow_binary: bool = False) -> None:
+    """Refuse binary input for the text-only tools unless explicitly overridden."""
+    if allow_binary:
+        return
+    kind = looks_binary(data)
+    if kind is None:
+        return
+    eprint(f"refusing to treat {origin} as text: it looks like {kind}.")
+    eprint("Use inspect_file.py / clean_file.py, which route by format,")
```
guillaumemeyer: **[suggestion]** `guard_binary()` always tells the user to “Use inspect_file.py / clean_file.py”. From `clean_file.py`’s new text fallback that is the tool they are already running — classify already decided the bytes were not a known container, so the advice is circular. Same when `inspect_file.py` hits an unknown binary.

**Suggestion:** Keep the generic message for the text-only scripts. From `clean_file.py` / `inspect_file.py`, say the bytes are not a supported text/image/container format and point at `--force-text` (or `--as …`) instead.

guillaumemeyer (dismissed):

skills/remove-ai-marks/scripts/clean_file.py
blackorange-mega: Confirmed and fixed in 7c59b8c. The sniff now runs before backup_path(), so the
refusal happens before any sidecar exists. Reproduced the original: exit 2 but
mystery.bin.bak left on disk.

The same hole applied to `--as text` on a .docx, which skips classify()
entirely. Both paths are covered now:
test_clean_file_in_place_refuses_before_taking_a_backup and
test_clean_file_in_place_as_text_on_docx_leaves_no_backup. Both assert the
input is byte-identical and the directory contains nothing but the input.

skills/remove-ai-marks/scripts/common.py
blackorange-mega: Right, and the failure mode is worse than codec drift. The text layer also does
universal-newline translation, so PNG's \x89PNG\r\n\x1a\n arrived as
\x89PNG\n\x1a\n and the magic never matched — it was refused by the NUL-byte
fallback instead, i.e. the detection was accidental and the reported format was
wrong. A container whose magic spans CR/LF and whose head carries no NULs would
have slipped past.

_read_stdin_capped now reads sys.stdin.buffer and guards the raw octets, with a
text fallback for a replaced stdin. Tests parametrise PNG and JPEG across
default, cp1252 and latin-1 stdio codecs; as you noted, the old ZIP-only test
could not catch this because "PK" is ASCII.

_configure_stdio() is untouched and stdout/stderr still go through it. Checked
that reading .buffer after reconfigure() loses nothing: 1.74 MB through the
multi-chunk path round-trips exactly.

skills/remove-ai-marks/scripts/common.py
blackorange-mega: Fixed. The advice is a parameter now: TEXT_TOOL_ADVICE keeps the pointer to the
routers for the text-only scripts, ROUTER_ADVICE is used by clean_file.py and
inspect_file.py and says the bytes match no supported text/image/container
format, pointing at --force-text / --as.
test_router_advice_is_not_circular asserts the routers never emit
"Use inspect_file.py / clean_file.py", and
test_text_only_scripts_keep_the_pointer_to_the_routers asserts the text tools
still do.

blackorange-mega:
Thanks for the review — all three addressed in 7c59b8c, each reproduced first.

- --in-place took the backup before the sniff; moved the sniff ahead of
  backup_path(). Also covers --as text on a .docx, which bypasses classify().
- stdin sniffed after decoding. Now reads sys.stdin.buffer. Beyond the codec
  issue you flagged, the text layer's newline translation was breaking PNG's
  magic outright, so it was only being caught by the NUL heuristic.
- guard_binary's advice is now a parameter, so the routers no longer point at
  themselves.

Also confirmed your merge resolution holds: the binary guard, the Windows stdio
reconfigure and --strip-emoji-glue are all intact — I ran the Windows CI smoke
steps from ci.yml locally on Windows and they pass.

132 tests passing. The new tests were checked against the pre-fix source and
all six fail there.

guillaumemeyer:
@blackorange-mega dunno why the test CI is blocked... I'll check later.
Thanks for the contribution.

guillaumemeyer (approved):

PR #25, opened 2026-08-14
fix: markdown frontmatter cleaner crashes and leaks nested AI keys

### LINKED ISSUE #26: [bug] clean_markdown() crashes on blank lines and leaves dropped keys' nested values behind
## What happened

`clean_markdown()` in `container_meta.py` has two problems sitting in the same loop.

First, it crashes on any frontmatter that has a blank line in it:

```python
>>> clean_markdown("---\ntitle: Demo\n\nauthor: you\n---\nBody\n")
IndexError: string index out of range
```

It runs `line[0]` on the blank line.

Second, when it drops a top-level AI key it keeps that key's nested lines, so the value it says it removed is still in the file:

```
actions: ['drop frontmatter key: model']
output:
title: Demo
  name: claude-opus
  version: 4
author: you
```

That also leaves the frontmatter as invalid YAML.

## What you expected

Frontmatter with blank lines should clean without crashing, and dropping a key should take its nested lines with it so nothing AI-ish is left behind.

## Steps to reproduce

1. import `clean_markdown` from `container_meta`
2. call it on a frontmatter doc with a blank line in it, or one with a nested AI key like `model:` and indented children under it
3. first one crashes, second one leaves the value behind

## Environment

- OS and arch: Windows 10 x64 (it's a logic bug though, repros anywhere)
- Python version: 3.14
- How you run the skill: scripts only
- Optional tools present: n/a

## Input type

- [x] Text (`.md`)
- Layer involved: Files (frontmatter metadata)

## Extra context

Opened a fix in #25.

poorvith-mp:
Fixes #26

`clean_markdown()` has two bugs living in the same loop.

**It crashes.** Any markdown whose YAML frontmatter has a blank line in it throws `IndexError`, because the loop runs `line[0]` on an empty string:

```
>>> clean_markdown("---\ntitle: Demo\n\nauthor: you\n---\nBody\n")
IndexError: string index out of range
```

**It leaks the thing it says it removed.** When a top-level key gets dropped, its nested lines are still kept, since nothing tracks that the parent was dropped. So dropping `model:` leaves `name: claude-opus` behind and the frontmatter ends up as invalid YAML:

```
actions: ['drop frontmatter key: model']
output:
title: Demo
  name: claude-opus     <- reported as removed, still there
  version: 4
author: you
```

The comment already in the code ("keep nested only if we kept a parent; track with flag") describes a flag that was never actually added. This adds it.

Why it matters: the tool tells you it stripped the provenance and then doesn't. `references/ethics.md` says reports shouldn't imply a clean happened when it didn't, so this is that exact case.

### changes

- `container_meta.py`: rewrote the `clean_markdown()` loop with a `dropping` flag, and guard blank/comment lines before indexing them.
- `tests/test_container_meta.py`: 4 regression tests (blank-line crash, nested-key leak, inspect round-trip, comment + list preservation).

`inspect_markdown()` / `_parse_simple_yaml_keys()` are left alone. They already guard `line[0]` with an early `continue`, so they never hit this.

### behaviour change

A blank line between a dropped key and the next kept key now gets dropped along with that block. Cosmetic only.

### checklist

- [x] matches `SKILL.md` / `references/removal-matrix.md`
- [x] tests added under `tests/`
- [x] `python3 -m pytest -q` passes (136 tests)
- [x] no docs change needed (bug fix, no new flags)
- [x] no unrelated refactors

### DIFF
--- a/skills/remove-ai-marks/scripts/container_meta.py
+++ b/skills/remove-ai-marks/scripts/container_meta.py
@@ -186,31 +186,43 @@ def clean_markdown(text: str) -> tuple[str, list[str]]:
     block = m.group(1)
     body = text[m.end() :]
     kept: list[str] = []
+    dropping = False  # inside the nested block of a dropped top-level key
     for line in block.splitlines():
-        if not line.strip() or line.strip().startswith("#") or line[0] in (" ", "\t", "-"):
-            # drop nested lines only if previous key was dropped — simple approach:
-            # keep nested only if we kept a parent; track with flag
-            if kept and line[0] in (" ", "\t", "-"):
-                kept.append(line)
-            elif not line.strip() or line.strip().startswith("#"):
+        stripped = line.strip()
+
+        # Blank lines and comments belong to whichever block we are inside.
+        if not stripped or stripped.startswith("#"):
+            if not dropping:
                 kept.append(line)
             continue
+
+        # Continuation lines (nested mappings, list items) follow their parent.
+        if line[0] in (" ", "\t", "-"):
+            if not dropping:
+                kept.append(line)
+            continue
+
         km = re.match(r"^([A-Za-z0-9_.-]+)\s*:", line)
-        if km:
-            key = km.group(1)
-            if key.lower() in AI_FRONTMATTER_KEYS or AI_META_NAME_RE.search(key):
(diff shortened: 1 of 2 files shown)


guillaumemeyer (approved):
Thanks for your contribution @prvthmpcypher

PR #28, opened 2026-08-14
feat: preserve script joiners, flag emoji, and Arabic Cf marks in Layer A

### LINKED ISSUE #27: [bug] Layer A strips script joiners (Persian/Indic), flag emoji tags, and Arabic Cf marks
## What happened

Layer A already preserves emoji ZWJ/VS glue now (nice), but a few other load-bearing invisible characters still get stripped and quietly change the text:

- Persian ZWNJ: `می‌روم` becomes `میروم`, which is a different word
- Devanagari conjunct: the ZWJ in `क्‍ष` gets dropped
- Flag emoji: `🏴󠁧󠁢󠁳󠁣󠁴󠁿` (Scotland) collapses to a plain `🏴` because the tag chars go
- Arabic/Syriac orthography marks like U+0600 and U+06DD (END OF AYAH) get removed as `other_cf`

repro on current `main`:

```python
>>> clean_text("x\u06DDy")   # ARABIC END OF AYAH
'xy'
>>> clean_text("می‌روم")
'میروم'
```

## What you expected

Same rule you already use for emoji glue: keep these when they're load-bearing, strip them when they're just carriers sitting between plain ASCII.

## Steps to reproduce

1. `clean_text` on any Persian/Indic word with a ZWNJ/ZWJ in it, a flag emoji, or an Arabic Cf mark
2. the invisible char gets removed and the visible text changes

## Environment

- OS and arch: Windows 10 x64 (logic bug, repros anywhere)
- Python version: 3.14
- How you run the skill: scripts only

## Input type

- [x] Text
- Layer involved: A (Unicode)

## Extra context

Kept it in the same shape as the existing emoji-glue handling: preserve ZWNJ/ZWJ when a neighbour is a complex-script letter, preserve flag tag chars after an emoji base, and allowlist the handful of Arabic/Syriac Cf marks. `--strip-emoji-glue` still strips everything for paranoid mode. Fix coming in a PR.

poorvith-mp:
Fixes #27

Layer A already preserves emoji ZWJ/VS glue after an emoji base, which is great. But a few other invisible characters that carry real meaning still got stripped, so cleaning text in some languages quietly changed the words.

Broken on `main`:

```python
>>> clean_text("x\u06DDy")   # ARABIC END OF AYAH
'xy'
>>> clean_text("می‌روم")     # Persian, ZWNJ
'میروم'                      # different word now
```

Same for Devanagari conjuncts (the ZWJ in `क्‍ष`) and flag emoji (`🏴󠁧󠁢󠁳󠁣󠁴󠁿` collapses to `🏴` once the tag chars go).

### what I did

Reused the machinery that's already there for emoji glue instead of adding a new system:

- ZWNJ/ZWJ (`U+200C`/`U+200D`) are kept when the previous kept char is a complex-script letter or mark (Arabic, Devanagari, etc). Between plain ASCII they're still carriers, so they still get stripped.
- Flag tag chars (`U+E0020`–`U+E007F`) are kept when they follow an emoji base, same idea as emoji glue.
- A small allowlist of orthographic Arabic/Syriac `Cf` marks (`U+0600`–`U+0605`, `U+06DD`, `U+070F`, `U+08E2`, `U+110BD`, `U+110CD`) is preserved instead of being dropped as `other_cf`.

`--strip-emoji-glue` (paranoid mode) still strips every one of these, so nothing changes for anyone who wants the old blanket behaviour.

### heads up

A ZWJ sitting between two non-ASCII letters that don't actually use it (say two random CJK chars) would now be preserved. That's a deliberate trade so real script text survives, and paranoid mode still removes it. Noted it in the code.

### changes

- `text_unicode.py`: extend `_decide()` with the three keep-rules, broaden the "don't advance prev_kept" guard from emoji glue to all glue.
- `clean_text.py` / `inspect_text.py`: updated `--strip-emoji-glue` help text.
- `references/mark-classes.md`: documented the preserve-by-default behaviour.
- `tests/test_clean_text.py`: preserve tests (Persian, Devanagari, flag, Arabic Cf), still-strips-between-ASCII tests, and a paranoid-mode test.

### checklist

- [x] matches `SKILL.md` / `references/mark-classes.md`
- [x] tests added under `tests/`
- [x] `python3 -m pytest -q` passes (141 tests)
- [x] docs updated (`mark-classes.md` + CLI help)
- [x] no unrelated refactors

### DIFF
--- a/skills/remove-ai-marks/references/mark-classes.md
+++ b/skills/remove-ai-marks/references/mark-classes.md
@@ -15,6 +15,8 @@ Invisible or near-invisible characters, exotic spaces, bidi controls, tag charac
 
 **Removal:** `clean_text.py` / Layer A — deterministic, verifiable.
 
+Load-bearing invisibles are preserved by default so real text is not corrupted: emoji glue (ZWJ/VS after an emoji base), script joiners (ZWNJ/ZWJ inside complex scripts like Persian or Devanagari), flag tag-char sequences, and orthographic Arabic/Syriac `Cf` marks. The same characters between plain ASCII stay carriers and are still stripped. Use `--strip-emoji-glue` for paranoid mode (strips all of them).
+
 Maps to Nature paper “edit-based watermarking.”
 
 ## 2. Generative / statistical text (token sampling)
--- a/skills/remove-ai-marks/scripts/clean_text.py
+++ b/skills/remove-ai-marks/scripts/clean_text.py
@@ -32,7 +32,7 @@ def main() -> int:
     p.add_argument(
         "--strip-emoji-glue",
         action="store_true",
-        help="Strip emoji presentation selectors/ZWJ even after an emoji base (paranoid)",
+        help="Paranoid: strip all load-bearing invisibles too (emoji glue, script joiners, flag tags, orthographic Cf)",
     )
     p.add_argument("--stats", action="store_true", help="Print stats JSON to stderr")
     p.add_argument(
--- a/skills/remove-ai-marks/scripts/inspect_text.py
+++ b/skills/remove-ai-marks/scripts/inspect_text.py
@@ -26,7 +26,7 @@ def main() -> int:
     p.add_argument(
         "--strip-emoji-glue",
         action="store_true",
-        help="Flag emoji presentation selectors/ZWJ even after an emoji base (paranoid)",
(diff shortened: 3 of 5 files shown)


guillaumemeyer (approved):
Thanks for your contribution @prvthmpcypher

PR #33, opened 2026-08-14
fix: print filename at top of inspect_file.py output

### LINKED ISSUE #31: [feature] Print the filename where the watermark was found
## Problem

If I were to run `inspect_file` on an entire codebase with something like:

`find "/path/to/codebase" -type f.....<more parameters>.... -exec python skills/remove-ai-marks/scripts/inspect_file.py --aggressive {} \; >results.txt`

I'd get a file full of potential results, but no indication as to which filename actually had the suspicious characters.

## Proposal

Add the filename to the top of the results, such as:
```

File: /path/to/codebase/ai_watermarked_file.py
Kind: text
Length: 29523 chars
Suspicious: 1
Hits:
  [space/informational] U+3000 IDEOGRAPHIC SPACE (Zs) x2 @ [1784, 2162]
Note: Layer A only: invisible/format Unicode and space homoglyphs (edit-based carriers).
Note: Statistical (token-sampling) watermarks are not detectable here; use Layer B rewrite.
Note: Inspect kinds: strip, bidi, tag_chars, variation_selector, zwj_family, space, confusable, other_cf.
Note: Emoji presentation glue (ZWJ/VS15/VS16 after an emoji base) is preserved by default; use --strip-emoji-glue for paranoid mode.
```

## Alternatives

N/A

## Extra context

N/A

poorvith-mp:
Closes #31.

## Problem

Running `inspect_file.py` over a whole codebase with `find ... -exec inspect_file.py {} \; > results.txt` produces one block of findings per file, but nothing in the output says which file a block belongs to. You end up with a wall of `Kind: text` / `Suspicious: N` blocks and no way to trace a hit back to its source.

## Fix

`inspect_file.py` now prints `Path: <file>` as the first line of output, before `Kind:`, for all three inspect kinds (text, image, container). Image and container already carried a path internally (from `report.path`), just printed in the wrong order; text didn't carry one at all since `inspect_text()` works on raw strings and has no concept of a file. Kept `text_unicode.py` untouched (it's also used by `inspect_text.py` on stdin, which has no path to report) and added the path at the `inspect_file.py` layer instead, where the file argument is known.

`--json` output for the text kind now also includes a `"path"` key, matching what image/container already emit.

Before:
```
Kind: text
Length: 29523 chars
Suspicious: 1
...
```

After:
```
Path: /path/to/codebase/ai_watermarked_file.py
Kind: text
Length: 29523 chars
Suspicious: 1
...
```

## Testing

- Added `tests/test_inspect_file_path.py` covering the text and container kinds (human output line order + JSON `path` key).
- `python -m pytest tests/` — 152 passed, 0 failed.

### DIFF
--- a/skills/remove-ai-marks/scripts/inspect_file.py
+++ b/skills/remove-ai-marks/scripts/inspect_file.py
@@ -81,8 +81,9 @@ def main() -> int:
         )
         report = inspect_text(text, aggressive=args.aggressive)
         if args.json:
-            emit_json({"kind": "text", **report.to_dict()})
+            emit_json({"kind": "text", "path": str(args.path), **report.to_dict()})
         else:
+            print(f"Path: {args.path}")
             print(f"Kind: text")
             print(human_report(report))
         return 0 if report.suspicious_total == 0 else 1
@@ -92,8 +93,8 @@ def main() -> int:
         if args.json:
             emit_json({"kind": "image", **report.to_dict()})
         else:
-            print(f"Kind: image")
             print(f"Path: {report.path}")
+            print(f"Kind: image")
             print(f"Format: {report.format}")
             print(f"C2PA: {report.has_c2pa}")
             print(f"AI metadata: {report.has_ai_metadata}")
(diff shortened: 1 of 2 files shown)


guillaumemeyer:
Superseded by merged #50 (same issue #31). Conflicting, weaker impl.

PR #43, opened 2026-08-14
feat(gui): add a desktop app for scanning and cleaning files (1/5)

blackorange-mega:
Replaces #41, which put the whole GUI in one diff. Same code, split into five reviewable pieces — this is the first, and each later one leaves a working app, so you can merge as few or as many as you want.

## What this adds

```bash
python3 gui/launch.py
```

Drag files in or press **Browse**, read the verdict, click clean, download the result. Handles PNG, JPEG, PDF, DOCX, ODT, SVG, HTML, Markdown and plain text, one at a time or in a batch.

Originals are never touched: files are copied into a temp directory that is removed when the app exits.

## How it fits the existing code

- `gui/server/bridge.py` calls the `skills/remove-ai-marks` scripts in-process — `inspect_file`, `clean_file`, `inspect_text`, `clean_text`. No cleaning logic is duplicated, so the GUI cannot drift from CLI behaviour.
- `gui/server/app.py` serves the page and a small JSON API. Assets come from a table built at startup, so a request path is never joined onto a directory; every `/api/` call needs a session token minted at launch, which keeps other pages in the same browser from driving a server that reads and writes local files.
- `gui/server/compat.py` works around `preexec_fn` being POSIX-only. The skill scripts pass it to `subprocess`, which made every `exiftool`/`c2patool` call raise on Windows — the tools looked missing when they were installed.
- `gui/web/` shows the findings, including the exact hidden characters in their surrounding text.

## Notes

- Python 3.10+, **stdlib only**. No new dependencies, nothing added to `requirements-dev.txt`.
- Nothing under `skills/` changes; `tests/test_gui_server.py` is added, covering path traversal, the asset table, header injection and the token guard.
- `python3 -m pytest -q` passes (156).
- Verified by hand in a browser: upload → inspect → clean → download, on a file carrying ZWSP, SHY and NBSP.

## The rest of the series

Branches are ready and rebase onto each merge; I'll open each one as its predecessor lands.

| | Slice | Adds |
| --- | --- | --- |
| 1 | this PR | Files screen — scan, clean, download |
| 2 | `feat/gui-text` | Text screen — paste text, reveal hidden characters in place |
| 3 | `feat/gui-rewrite` | Rewrite screen — Layer B, prompt-only by default |
| 4 | `feat/gui-setup-guide` | Setup (tool diagnostics) and Guide (your reference notes, offline) |
| 5 | `feat/gui-native-dialogs` | Native file dialogs, save next to or over the original, double-click launchers |

Happy to reorder, drop any of them, or keep the whole thing under `gui/` out of the default install path if you'd rather it stay optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,18 @@ Vendors / ecosystems (class-level): **Claude**, **Gemini / SynthID-Text**, **Ope
 Skill path: [`skills/remove-ai-marks/`](skills/remove-ai-marks/)  
 (migration: formerly `remove-claude-marks`; slash alias `/remove-claude-marks` still documented)
 
+## Desktop app (no terminal needed)
+
+```bash
+python3 gui/launch.py
+```
+
+Drag files in, read the verdict, click clean. Paste text to see hidden characters revealed in
+place. Python 3.10+ stdlib only — no packages to install. Runs on **Windows, macOS and Linux**,
+loopback-only, with every action mapped to the same scripts the CLI calls.
+
+Details: [`gui/README.md`](gui/README.md)
+
 ## Install (agent skill)
 
 ```bash
--- a/gui/README.md
+++ b/gui/README.md
@@ -1,0 +1,127 @@
+# watermarks-remover — desktop app
+
+A point-and-click front end for the `remove-ai-marks` scripts. No terminal, no
+Python packages to install, and nothing leaves your computer.
+
+Works on **Windows, macOS and Linux**.
+
+---
+
+## Start it
+
+```bash
+python3 gui/launch.py
+```
+
+The app opens in its own window if you have [pywebview](https://pywebview.flowrl.com/)
+installed, and in your default browser otherwise. Both are the same app — the
+page is served from your own machine on `127.0.0.1`.
+
+**Requirements:** Python 3.10 or newer, and nothing else. If Windows says Python
+is missing, install it from [python.org](https://www.python.org/downloads/) and
+tick *“Add python.exe to PATH”* during setup.
+
+Optional native window:
+
+```bash
+pip install pywebview
+```
+
+---
+
+## The Files screen
+
+Drag files in, or press **Browse…**. Every file is scanned as it arrives and
+gets a verdict: *marks* or *clean*. Open one to see the findings, the exact
+hidden characters it contains, and where they sit in the text.
+
+**Clean this file** strips the marks and hands back a copy to download —
+`draft.md` → `draft.cleaned.md`. The original is never touched.
+
+Handles PNG, JPEG, PDF, DOCX, ODT, SVG, HTML, Markdown and plain text. Batch
+work is fine: **Clean all**, then **Download all** for a zip.
+
+| Option | Effect |
+| --- | --- |
+| Treat file as | Overrides format detection, same as the CLI's `--as` |
+| Flag look-alike letters | Reports Cyrillic/Greek letters posing as Latin ones |
+| Also replace those letters | Rewrites them during a clean (`--aggressive-homoglyphs`) |
+| Normalise text (NFKC) | Folds ﬁ, ①, full-width forms into plain equivalents |
+| Keep camera metadata | Images: drop only AI/C2PA segments, keep EXIF |
+| Score SynthID | Appears only when `REVERSE_SYNTHID_DIR` is configured |
+
+#### Word and ODT documents
+
+A `.docx`/`.odt` is a zip, and its prose lives in `word/document.xml` /
+`content.xml`. Two consequences worth knowing:
+
+- **`inspect_text.py` must not be pointed at one.** It reads the file as text,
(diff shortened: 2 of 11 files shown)


guillaumemeyer:
@blackorange-mega I really like the idea of a web UI.
I'm still wondering whether it should be part of this repo or deserve its own dedicated one.

Option 1 - Same repo:
- Pro: single versioning/discovery — CLI and UI never drift apart; shared LICENSE, SECURITY.md, CI, and Makefile patterns (you already have docker-*-build targets a UI could ride on).
- Pro: the web layer would shell out to the same skills/remove-ai-marks/scripts/* you test today, so one suite covers both.
- Con: this repo is deliberately stdlib-only, zero-dependency, CLI/agent-first (CONTRIBUTING.md:26, README:53). A web UI means a framework, JS assets, and a long-running server — it breaks that philosophy on day one.
- Con: a server that accepts arbitrary files is a much bigger security surface. Your recent hardening (safe atomic writes, loopback-only HTTP, resource caps) is about CLI safety; a web endpoint turns every hardening decision into an auth/upload/job-management problem.
- Con: heavy backends (CtrlRegen ~10GB models, torch) are long-running batch jobs — they map poorly onto a request/response web UI and would bloat this lean repo.

Option 2 - Separate repo:
- Pro: web UI gets its own threat model (auth, upload limits, job queue) without forcing the CLI repo to carry web concerns.
- Pro: UI can iterate on its own cadence; the CLI repo stays dependency-free and reviewable.
- Con: duplication — CI, docs, release train, security reporting all fork; syncing them is real overhead for a single maintainer.
- Con: drift risk: the UI shells out to scripts that are moving fast; you'll need a pinned-release contract or a shared package.

I am leaning toward option 2 for now...

blackorange-mega:
Thanks for writing this out properly instead of just closing it. Fair question to be asking, too.

Let me push back on a couple of the cons — not to argue you out of option 2, but because I think two of them describe a different program than the one in these branches.

On dependencies: there's no framework and no build step. `gui/` imports the standard library and your own scripts under `skills/remove-ai-marks/scripts`, and that's the whole list. `requirements-dev.txt`, the Makefile, the Dockerfiles and CI are untouched across all five PRs. The front end is three files — `index.html`, `app.css`, `app.js`, about 96 KB together — with no npm, no bundler, no CDN. The only URL anywhere in them is `127.0.0.1`. `pywebview` is the one optional import and it sits in a try/except; without it you just get a browser tab instead of a native window.

On the security surface: it isn't really a web app, it's a local process that happens to draw its UI in a browser. It binds `127.0.0.1` on a random port, every `/api/` call needs a session token minted at launch and compared with `compare_digest`, Host and Origin get checked, responses carry CSP and `nosniff`, bodies are capped at the existing `MAX_INPUT_BYTES`, uploads go to a temp dir that's deleted on exit, and static files come out of a table built at startup rather than from joining a request path onto a directory. No auth, no quotas, no job queue — because nothing is on the network and there's exactly one user, whoever started it.

You're still right that it's more surface than a CLI, though, and I'd rather admit that with a receipt than talk around it: your CodeQL gate caught four things in my first version, path handling and a response header. They're fixed and it's green now, but a CLI would never have had those bugs to begin with, and that asymmetry doesn't go away.

The heavy backends one just doesn't apply here. Nothing under `gui/` touches CtrlRegen or torch. It does Layer A and file metadata, which come back fast, and SynthID stays behind `REVERSE_SYNTHID_DIR` if you want it at all. If this ever grew into driving 10 GB models it would need a real job queue, but that's a different program and I wouldn't put that one here either.

All that said — option 2 is genuinely fine by me. It already works, too: `bridge.py` reads `WATERMARKS_SKILL_DIR`, so the UI runs against an outside checkout with no code changes, and I've got it at `blackorange-mega/watermarks-remover-gui` already.

The drift worry is the one I'd take seriously, and it's slightly worse than you framed it: the GUI imports your scripts as modules rather than shelling out, so it's coupled tighter than a CLI contract would be. Across two repos I'd pin to your release tags and treat the modules it uses — `text_unicode`, `common`, `image_meta`, `container_meta`, `rewrite_text` — as the interface. If you'd rather I shell out to the documented CLI instead, I'll convert it. Cleaner boundary for you, a bit slower for me, and worth it if it means you never have to think about my repo when you refactor yours.

So: want me to close #43–#47? No hard feelings if so, none of it is wasted if it lives elsewhere. And if it's any use I can leave a one-line pointer in your README so people can find the UI — or leave it out entirely if you'd rather keep this repo CLI-first.

guillaumemeyer:
We talked with other contributors, and the consensus is to create separate repos for GUIs, then register them as part of the ecosystem in the README.
See https://github.com/guillaumemeyer/watermarks-remover/pull/71 for reference.
If you agree, would you be okay with me closing the related PRs, and I'll promote the registered apps in the ecosystem?

PR #70, opened 2026-08-15
fix: macOS portability — pure --json stdout for the SynthID scorer, BSD realpath probe

adrielw:
Two small portability fixes found running v0.5.0 on macOS.

### 1. `score_synthid.py --json` stdout is corrupted by upstream prints

The reverse-SynthID upstream prints progress (`CodebookV4 loaded: ...`)
straight to stdout. `image_meta.py` parses the scorer's stdout with
`json.loads`, so every score comes back as
`{"available": false, "error": "bad scorer JSON: ..."}` whenever the
backend is configured.

Fix: redirect stdout to stderr around the upstream calls so `--json` owns
stdout. Added `tests/test_synthid_stdout_purity.py`, which drives the real
script against a deliberately noisy stub upstream (including a stub `cv2`,
so it runs without OpenCV). The test fails on current main and passes with
the fix.

### 2. `setup_synthid.sh` / `setup_ctrlregen.sh` abort on macOS

`command -v realpath` is true on macOS, but BSD realpath has no `-m`, so
both bootstraps take the GNU branch and abort on the first path they
normalize. Probing the flag itself (`realpath -m .`) routes BSD systems to
the portable `pwd` fallback that is already there.

Verified on macOS (Darwin 25.6.0): the probe now selects the fallback and
both scripts pass `bash -n`. Full suite: 228 passed, 1 skipped.


guillaumemeyer (approved):

guillaumemeyer:
Merged @adrielw

Issue #77, opened 2026-08-15
[RFC / Proposal] Built-in lightweight single-file Web UI served by server.py


ivanusto:
Hi @guillaumemeyer,

First of all, thank you for building this fantastic project! The stdlib-only architecture and comprehensive handling of multi-vendor AI provenance marks are really well thought out.

I noticed that `server.py` already exposes a clean HTTP API (`/clean`, `/inspect`, `/health`). To make the tool accessible to non-technical end-users who might not use CLI or AI agent skills directly, I am considering adding a minimal, zero-dependency Web UI.

### Proposal
Serve a lightweight, single-file HTML/CSS/JS frontend directly at `GET /` via `server.py`:
- **Zero extra dependencies**: Plain HTML/CSS/vanilla JS embedded or served as a static file (adhering to the stdlib-only philosophy).
- **Core capabilities**:
  - Direct text paste & clean (with diff/statistics of stripped zero-width characters and homoglyphs).
  - Drag-and-drop file/image upload for metadata/C2PA stripping with instant download.
- **Client-side compatibility**: Providing an optional 100% client-side mode so users can also run it purely in-browser (e.g. GitHub Pages) with zero privacy/server concerns.

### Question for Maintainers
Would you be open to a PR that introduces this lightweight Web UI into `server.py` / the repository? 

Or do you prefer keeping the core repository strictly headless/CLI/Agent-focused and having Web UIs maintained as standalone external clients (which we could link to in the README)?

Looking forward to your thoughts! Thanks again for the great work.

ivanusto:
Following up on my own proposal after reading the discussion in #43 — I think you're right that a UI belongs in its own repo, and I've gone that route rather than push for an in-tree `server.py` UI.

**What I did instead**

- Moved the web UI to a standalone repo: **[ivanusto/watermarks-remover-web](https://github.com/ivanusto/watermarks-remover-web)** (MIT, upstream LICENSE preserved in `NOTICE`, clearly labelled as an independent client, not affiliated).
- The browser engines are straight ports of `text_unicode.py` and `image_meta.py` — including the load-bearing-invisible rules from #38/#52 and the `Cf` catch-all — with a parity test suite that runs the JS and the Python side by side against an upstream checkout (140 text cases byte-identical, PNG/JPEG/WebP strip output byte-identical). No canvas re-encode, no third-party requests, CSP on, nothing persisted unless the user opts in.
[...]

PR #86, opened 2026-08-16
docs: add unmark-web to Ecosystem (#77)

ivanusto:
Adds a short entry for [unmark-web](https://github.com/ivanusto/unmark-web) to the Ecosystem chapter, following the format of the existing MetaClean entry (factual, no endorsement claim), as discussed in #77. README only, no code changes.

The project was renamed from `watermarks-remover-web` to `unmark-web` per the request in this thread.


guillaumemeyer:
@ivanusto, if you don't mind, would you be okay with renaming your repo and the mention to something less similar to `watermarks-remover`?
The Hermes Agent community (I do not compare, ofc ;-) ) is struggling with many projects that include hermes-agent in their name, which makes it hard for users to tell what is what.

And no, I am not suggesting creating a "Watermarks Remover Foundation" like [Rust](https://www.reddit.com/r/rust/comments/12lb0am/can_someone_explain_to_me_whats_happening_with/) ;-)

ivanusto:
Absolutely — that's a fair ask, and much better to fix now than after people start linking to it.

Done: the project is now **unmark-web** — https://github.com/ivanusto/unmark-web (GitHub redirects the old URL; the demo moved to https://ivanusto.github.io/unmark-web/). I renamed the in-app title as well, not just the repo, so nothing about it reads as an official component of this project.

I also pushed an update that takes the Copilot review into account: the entry no longer names your internal modules, my proxy script, or a list of formats — the parts most likely to drift. It now reads:

> ### unmark-web — browser web UI
>
> [unmark-web](https://github.com/ivanusto/unmark-web) is an independent, MIT-licensed static web client. It removes invisible Unicode marks from text and strips provenance metadata from images entirely in the browser, and can optionally call this repository's HTTP service for the formats it does not handle locally. It is a separate codebase and is not affiliated with this project; see its README for scope and limits.

If you'd prefer a different name, or want the entry trimmed further, just say so and I'll adjust.

One small suggestion, if it's useful: a line under "Adding a project" — something like "please avoid names that start with or closely resemble `watermarks-remover`" — would set the expectation up front for future entries instead of case by case. Happy to open a separate PR for that if you want it.

guillaumemeyer (approved):

guillaumemeyer:
@ivanusto, thanks for renaming the project. Pushing a PR now with the suggested README update.

guillaumemeyer:
@ivanusto if promoting this on linkedin/x/fb/insta, please mention me so I can share

ivanusto:
Thanks for merging, and for turning the naming point into #108 rather than leaving it as a one-off ask — having it written down under "Adding a project" is better than the two of us settling it in a thread.

Noted on the social side: no immediate plans, but if I write anything up about it I'll tag you.

One thing on my end, so the Ecosystem entry doesn't quietly go stale: I saw #107 land. unmark-web is a port of your `text_unicode.py` and `image_meta.py` with a daily hash check against them, so BMP/GIF/TIFF/EPUB are on my list to follow. The entry deliberately avoids listing formats, so it stays accurate either way — but I'd rather the port not drift far behind.

PR #35, opened 2026-08-14
feat: add a lightweight Cursor text skill

Zhaohan-Wang:
## Summary
- add a focused, self-contained `clean-user-facing-text` skill for Cursor
- preserve code, facts, citations, required disclosures, and multilingual formatting while applying text hygiene to authorized prose
- provide a cross-platform Cursor installer with staged replacement and backups
- include an optional Cursor Rule, concise installation documentation, and isolated installer tests

## Scope and safety
This revision intentionally removes Codex-specific installation, metadata, instructions, and tests. The package is text-only and omits image, C2PA, service, and external-model tooling. Deterministic Unicode cleanup remains auditable; statistical watermark reduction is described only as best-effort and never as proof of human authorship or guaranteed detector evasion.

The lightweight runtime is synchronized with the latest upstream text scripts and remains independently installable without the HTTP service.

## Test plan
- [x] Merge the latest `main` and resolve README/Makefile conflicts
- [x] Run the complete test suite (`232 passed, 1 skipped`)
- [x] Validate the Cursor skill structure
- [x] Compile the Python installer and validate the Bash wrapper
- [x] Test install, existing-install protection, forced replacement, and backup behavior

PR #89, opened 2026-08-16
Add Codex and Claude installation for remove-ai-marks

vesam918:
## What changed

- extend `install_skill.py` to install either bundled skill for Codex, Claude, Cursor, or every supported agent
- add Codex-facing `agents/openai.yaml` metadata for `remove-ai-marks`
- preserve the existing Cursor installer defaults and atomic backup behavior
- replace GNU-only `base64 -w0` examples with macOS/Linux-compatible commands
- document the new installation commands and add installer coverage

## Why

The full `remove-ai-marks` skill was documented only for Grok, while the installer handled only the lightweight Cursor skill. This makes the same agent-agnostic `SKILL.md` directly installable in Codex and Claude without duplicating the skill.

## User impact

Users can install the full skill with one command into `~/.codex/skills`, `~/.claude/skills`, or all supported agent homes. Existing installations remain untouched unless `--force` is passed.

## Validation

- `uv run --with-requirements requirements-dev.txt python -m pytest` — 285 passed, 1 skipped
- `quick_validate.py skills/remove-ai-marks` — valid
- focused installer tests cover Codex, Claude, all targets, backups, and metadata

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 !/docs/**
 !/service/
 !/service/**
+!/skills/
 !/skills/remove-ai-marks/
 !/skills/remove-ai-marks/**
 !/tests/
--- a/README.md
+++ b/README.md
@@ -34,6 +34,15 @@ Service path: [`service/`](service/)
 The skill ships **no code** — it calls the service over HTTP. Install the skill (markdown only) and start the service, then set `WATERMARKS_SERVICE_URL` if it is not `http://127.0.0.1:8765`.
 
 ```bash
+# Codex (~/.codex/skills/remove-ai-marks)
+python3 install_skill.py --target codex --skill remove-ai-marks
+
+# Claude (~/.claude/skills/remove-ai-marks)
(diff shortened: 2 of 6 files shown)


guillaumemeyer:
I think adding this would confuse some users and lead them to believe it's a good idea to use this skill within Claude, Claude Code, or Codex, which apply text watermarks.
See this conversation: https://github.com/guillaumemeyer/watermarks-remover/pull/35
Am I misinterpreting this PR's goal @vesam918 ?
Happy to discuss that

poorvith-mp (commented):
+1 to Guillaume here.

As discussed in #35, we intentionally don't provide Codex or Claude skill installers because both backends apply text watermarking during generation—using them to run Layer B rewrites is circular.

The macOS `base64` portability fix (`base64 < file | tr -d '\r\n'` instead of `-w0`) in `SKILL.md` is a great standalone improvement though. If you'd like to open a small PR with just the base64 portability update, happy to review and merge that!

PR #88, opened 2026-08-16
feat: recursively inspect and clean embedded raster data URIs in SVGs, HTML, and Markdown

poorvith-mp:
## What

Adds recursive inspection and cleaning of embedded raster and vector data URIs (`data:image/...`) inside SVG, HTML, and Markdown pipelines:

- **Scanner (`_inspect_embedded_data_uris`)**: Detects base64 and URL-encoded image data URIs (`png`, `jpeg`, `webp`, `avif`, `heic`, `svg+xml`) inside documents and flags embedded C2PA manifests, APP11 JUMBF chunks, and AI provenance markers.
- **Stripper (`_clean_embedded_data_uris`)**: Decodes embedded payloads, runs native image strippers (`strip_png`, `strip_jpeg`, `strip_webp`, `strip_isobmff`, or recursive `clean_svg`), and re-embeds cleaned payloads in-place.
- **Safeguards against edge cases**:
  - **Whitespace resilience**: Strips internal whitespace / line breaks from multiline base64 payloads prior to decoding (common in formatted SVG/CSS).
  - **Zero DOM churn**: If an embedded image is already clean (`cleaned_bytes == raw_bytes`), the original attribute is left 100% untouched.
  - **Malformed input protection**: Gracefully catches decode errors on corrupt data URIs and leaves the original string intact without failing the document cleaner.
  - **Nested SVG support**: Recursively cleans nested SVG payloads in data URIs.

## Why

Fixes #87. Plugs an untouched watermark bypass vector where AI generation tools (Adobe Firefly, Midjourney, Figma, DALL-E) embed raster images with full C2PA and EXIF metadata directly into vectors and webpages.

## Checklist

- [x] Behaviour matches `SKILL.md` / `references/removal-matrix.md` when relevant
- [x] Unit tests updated or added under `tests/` (`test_embedded_data_uris.py` added with 7 tests, 283 total passing)
- [x] `python3 -m pytest -q` passes
- [x] Docs updated (README and/or skill references) if user-facing behaviour changes
- [x] No drive-by refactors unrelated to the fix or feature

## Notes for the reviewer

- Tested with synthetic SVGs, HTML, and Markdown containing embedded PNGs, JPEGs, and nested SVGs under `tests/test_embedded_data_uris.py`.
- 100% Python stdlib (`base64`, `re`, `urllib.parse`).

PR #92, opened 2026-08-16
Handle malformed DOCX/ODT beyond BadZipFile in container inspectors

eeshsaxena:
`inspect_docx`, `inspect_odt`, and `detect_container_format` already catch `zipfile.BadZipFile` and return a graceful "not a valid zip" result, so they clearly mean to tolerate a bad container. But that's only one of the ways an untrusted DOCX/ODT can be unreadable, and the rest escape:

```python
from container_meta import inspect_docx
inspect_docx(corrupt_docx_bytes)
# NotImplementedError: That compression method is not supported
# (also RuntimeError for encrypted, zlib.error for bad deflate,
#  EOFError for truncated, ValueError: negative seek value, ...)
```

Reading a member of a corrupt/truncated/encrypted/unsupported-compression zip raises any of `NotImplementedError`, `RuntimeError`, `zlib.error`, `EOFError`, `OSError`, or `ValueError`, none of which the inspectors caught. Since these run over untrusted uploads (and in the audit loops), I collected the family into a `_ZIP_PARSE_ERRORS` tuple and broadened the three catches, so any unparseable container degrades the same way a plain `BadZipFile` already does. Added a test that crafts an unsupported-method DOCX plus truncated/garbage inputs; it raises `NotImplementedError` without the change and passes with it.

### DIFF
--- a/service/scripts/container_meta.py
+++ b/service/scripts/container_meta.py
@@ -11,9 +11,36 @@ import re
 import subprocess
 import urllib.parse
 import zipfile
+import zlib
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
+
+
+class ZipBudgetExceeded(Exception):
+    """A zip's declared decompressed size exceeds the processing cap.
+
+    Kept separate from the parse errors below so a refused zip bomb keeps
+    propagating (as it already does out of the clean_* helpers) instead of
+    being reported as an unparseable container.
+    """
+
+
+# A corrupt, truncated, encrypted, or unsupported-compression zip surfaces as
+# more than just BadZipFile: reading a member can raise NotImplementedError
+# (unknown compression/version), RuntimeError (encrypted), zlib.error (bad
+# deflate stream), EOFError (truncated), OSError (invalid stream), or
+# ValueError (e.g. a negative seek).
+_ZIP_PARSE_ERRORS = (
+    zipfile.BadZipFile,
+    zipfile.LargeZipFile,
+    NotImplementedError,
+    RuntimeError,
+    EOFError,
+    OSError,
+    ValueError,
+    zlib.error,
+)
 
 from common import (
     classify_finding_confidence,
@@ -162,7 +189,7 @@ def detect_container_format(path: Path, data: bytes | None = None) -> str:
                         return "odt"
                     if "META-INF/container.xml" in names and any(n.endswith(".opf") for n in names):
                         return "epub"
-            except zipfile.BadZipFile:
+            except _ZIP_PARSE_ERRORS:
                 pass
     return "unknown"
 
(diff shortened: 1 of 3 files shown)


guillaumemeyer:
Hey @eeshsaxena 
Thanks for this. The bug is real, I verified it: without the change, the test dies with NotImplementedError: That compression method is not supported. The fix itself is small and focused. A few things before this can merge:

Rebase onto main. The PR is currently conflicting because #88 reformatted the import block in container_meta.py, right where _ZIP_PARSE_ERRORS gets inserted. It's a single hunk, so it should be quick.

In the test, assert the degraded results, not just "no exception". Right now test_container_inspectors_survive_malformed_zip only checks that nothing raises. The point of the change is graceful degradation, so please assert inspect_docx(data) == (False, False, ["not a valid DOCX zip"], {}) plus the ODT and "unknown" equivalents. That also pins the truncated and garbage cases, which already raised BadZipFile before this change.

Don't swallow the zip-bomb rejection. _check_zip_budget raises ValueError("zip decompressed size exceeds cap ...") on purpose, inside the same try block, and your broader except now reports that as "not a valid DOCX zip". A file that's too big is not an unparseable container, and clean_docx/clean_odt still propagate that same ValueError, so inspect and clean now disagree on the same file. Easiest fix: give the budget check its own exception (a small ZipBudgetExceeded subclass of Exception) and let it propagate, or handle it before the generic tuple.

Optional, since you'll be in the test anyway: the comment lists zlib.error, encrypted, and OSError cases, but none of them are exercised. A bad-deflate or encrypted entry would cover the new exception types.

eeshsaxena:
Thanks for the thorough review, all three points were fair.

Rebased onto main, the import-block conflict from #88 is gone.

The zip-bomb case was a real oversight, you're right that inspect and clean were disagreeing on the same file. `_check_zip_budget` now raises a dedicated `ZipBudgetExceeded` (subclass of `Exception`, not in the parse-error tuple), so a refused bomb propagates out of the inspectors the same way it already does out of `clean_docx`/`clean_odt`. I updated the two budget tests in `test_security_hardening.py` to expect it.

The inspector test now asserts the actual degraded results, `(False, False, ["not a valid DOCX zip"], {})` and the ODT/unknown equivalents, for both the truncated and garbage cases, plus a test that the budget rejection propagates from `inspect_docx`.

poorvith-mp (dismissed):
Solid hardening. Handling `RuntimeError` (encrypted archives), `NotImplementedError`, and `zlib.error` prevents unhandled crashes on corrupt or encrypted OOXML/ODT containers.

Good call separating `ZipBudgetExceeded` so decompressed size caps aren't swallowed by `_ZIP_PARSE_ERRORS`.

*(Note for follow-up: we can also apply `_ZIP_PARSE_ERRORS` to `inspect_epub()` which landed in #107).*

Approved!

guillaumemeyer (dismissed):

PR #94, opened 2026-08-16
Feat strip reserved default ignorable code points in layer a

PinoEire:
## What

U+2065, U+FFF0..U+FFF8, U+E0000, U+E0080..U+E00FF, and U+E01F0..U+E0FFF
are unassigned code points carrying Other_Default_Ignorable_Code_Point,
reserved for future default-ignorable characters. Conformant renderers
display them invisibly, normalisation preserves them, and the Cf
catch-all never sees them (they are category Cn), which made them ideal
covert carriers that both inspect_text and clean_text passed through
untouched.

## Why

Missed ignorables

## Checklist

 service/scripts/text_unicode.py, skills/clean-user-facing-text/scripts/text_unicode.py, tests/test_clean_text.py, tests/test_lightweight_skill.py, skills/remove-ai-marks/references/mark-classes.md, and README.md.


guillaumemeyer:
Thanks for this @PinoEire

Two small things before merge:
- The new reserved_ignorable kind isn't in the "Inspect kinds:" note that both engines print in the CLI report. The report now emits a kind it doesn't list, so add it to that string in both copies.
- The README says U+3164/U+FFA0 are treated "like their already-covered siblings (U+115F/U+1160)", but the siblings are preserved after a Hangul jamo while the new ones are stripped unconditionally. I verified: ᄀ + U+3164 + ᅡ loses the filler, ᄀ + U+115F + ᅡ doesn't. Either wire them into the context preservation too, or reword the entry to say they're always stripped.

PR #98, opened 2026-08-16
feat: self-contained skill bundle for Claude Desktop (sandboxed, serverless)

PinoEire:
## Why

Claude Desktop runs skills inside a sandboxed VM whose network policy blocks all routes to the host, so the thin-client skill can never reach a local service on `127.0.0.1:8765` (nor a LAN address). The existing HTTP skill is therefore unusable in Desktop.

## What

- **`package_desktop_skill.py`** (+ `make desktop-skill`) assembles `dist/remove-ai-marks-desktop.zip` at build time: a Desktop-specific `SKILL.md` (`skills/remove-ai-marks-desktop/`), the shared `references/`, and the stdlib engine scripts copied fresh from `service/scripts/`. Nothing is vendored, so there are no duplicate engine copies to drift (unlike the Cursor skill, this variant wants byte-identical service behaviour, and a test enforces it).
- **`skills/remove-ai-marks-desktop/SKILL.md`**: same workflow as the HTTP skill (inspect first, deterministic clean, always offer Layer B, honest reporting) but drives `scripts/inspect_file.py` / `scripts/clean_file.py` directly. Sandbox degradations are stated up front: best-effort PDF without `exiftool`/`qpdf`, Layer B via the agent-as-rewriter prompts only, no pixel/scoring backends.
- **Bundled scope**: text Layer A, image metadata (PNG/JPEG/WebP/AVIF/HEIC), containers (SVG/DOCX/ODT/HTML/MD, PDF best-effort), stylometry, directory audits, `rewrite_text.py` print-prompt backend. Excluded: `server.py`, `audit_website.py`, every optional heavy backend and setup script.
- **Tests** (`tests/test_desktop_skill.py`): exact zip member list (backend/server files can never leak in), frontmatter validity, byte-equality of bundled scripts against the service copies, an extract-and-run smoke proving the bundle is self-contained, and the CLI entry point. Line-ending agnostic for the Windows runner.
- **`.gitignore` fix**: git never descends into an excluded directory, so the deny-by-default `/*` rule silently swallowed any *new* file under `skills/` despite the per-skill negations. `/skills/` is now re-included first, its children re-excluded, then each skill negated, and `skills/clean-user-facing-text/` gains the explicit allow it was missing.

## Testing

`python -m pytest`: 284 passed, 3 skipped (the usual optional-tool skips) on Windows; the new tests avoid platform-specific assumptions.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,17 @@
 !/docs/**
 !/service/
 !/service/**
+# git never descends into an excluded directory, so /skills/ itself must be
+# re-included before per-skill rules can match anything beneath it.
+!/skills/
+/skills/*
+!/skills/clean-user-facing-text/
+!/skills/clean-user-facing-text/**
 !/skills/remove-ai-marks/
 !/skills/remove-ai-marks/**
+!/skills/remove-ai-marks-desktop/
+!/skills/remove-ai-marks-desktop/**
+!/package_desktop_skill.py
 !/tests/
 !/tests/**
 !/.dockerignore
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 	smoke-markllm bootstrap-markllm docker-markllm-build docker-markllm-help \
 	smoke-markdiffusion bootstrap-markdiffusion docker-markdiffusion-build docker-markdiffusion-help \
 	docker-core-build docker-core-help serve compose-up compose-up-heavy compose-check \
-	install-skill install-cursor-text-skill clean
+	install-skill install-cursor-text-skill desktop-skill clean
 
 SCRIPTS := service/scripts
 PYTHON ?= $(shell if [ -x .venv/bin/python ]; then echo .venv/bin/python; else echo python3; fi)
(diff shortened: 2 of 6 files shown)


guillaumemeyer:
I think adding this would confuse some users and lead them to believe it's a good idea to use this skill within Claude, Claude Code, or Codex, which apply text watermarks.
See this conversation: https://github.com/guillaumemeyer/watermarks-remover/pull/35
Am I misinterpreting this PR's goal @PinoEire ?
Happy to discuss that

poorvith-mp (commented):
+1 to Guillaume's point here.

Because Claude applies its own token-sampling watermarks (Claude seam) during generation, packaging this as a native Claude Desktop skill creates a circular trap where users ask Claude to remove watermarks, only for Claude's sampler to re-embed its own statistical watermarks into the output prose.

That’s the same reason Codex was scoped out in #35. For users wanting serverless / non-HTTP cleaning, running the CLI scripts directly (`python service/scripts/clean_file.py`) already provides pure-stdlib local execution without needing a Claude wrapper.

Unless there's a specific use case that avoids this circular sampling issue, I don't think we should ship a dedicated Claude Desktop skill. (Also noting merge conflicts with `main` in `Makefile` and `README.md`).

PR #109, opened 2026-08-17
feat: vendor text-watermark detection (Gemini SynthID, Claude seam, MarkLLM) + SynthID image scorer sidecar

guillaumemeyer:
## What

Adds Layer B watermark **detection** as a first-class service capability, alongside the existing cleaning paths.

- **\`text_detectors.py\`** — a registry of text-watermark detectors behind one interface:
  - \`gemini-synthid-text\`: Google's official SynthID-text detector via the Gemini API (\`taskType: DETECT_TEXT_WATERMARK\`), activated by \`WATERMARKS_GEMINI_API_KEY\`.
  - \`claude-text\`: placeholder for Anthropic's announced detection API (reports unavailable until it ships).
  - \`markllm\`: the existing MarkLLM research harness (KGW / SynthID), same-config-only.
  - Fail-soft: unconfigured / errored detectors never block cleaning.
- **\`server.py\`** — new \`POST /detect\` endpoint, \`detect_before\` / \`detect_after\` options on \`/clean\` (before/after scoring for text and images), an opt-in \`/inspect\` \`"detect": true\` flag, and \`/capabilities\` gains \`text_detectors\` + \`scorers.synthid_http\`.
- **\`synthid_score_server.py\`** — a stdlib HTTP sidecar for the reverse-SynthID scorer, so the published core image never bundles the non-commercial upstream code (wired via \`WATERMARKS_SYNTHID_SCORER_URL\`).
- **\`score_synthid.py\`** — extract \`score_file()\` shared by the CLI and the sidecar.
- **compose / Dockerfile / .env** — \`wr-synthid-score\` sidecar service and env wiring.
- **Docs + tests** — README + skill docs; tests for detectors, the \`/detect\` endpoint, and the image sidecar.

## Scope note

\`image_meta.py\`'s HTTP-scorer hook is already on \`main\` (part of #107), so it is intentionally not included here.

## Verification

- \`ruff check\` clean, \`ruff format --check\` clean
- 348 passed, 1 skipped
- OpenAPI spec validates; \`docker compose config\` valid

Issue #106, opened 2026-08-17
Add per-candidate MarkLLM scoring for Layer B rewrites


ai-kunkun:
## Problem

`rewrite_text.py` already supports generating multiple rewrite candidates via `--candidates N`, but candidate selection is currently based only on bigram Jaccard lexical divergence, with a small length-drift penalty.

When `--markllm-scheme` is enabled, MarkLLM verification runs on the original text and then only on the candidate that has already been selected. This makes it difficult to evaluate whether lexical divergence actually correlates with watermark removal.

For example, the most lexically divergent candidate may still be detected as watermarked, while a slightly less divergent candidate may already fall below the detector threshold.

This becomes particularly relevant when evaluating paraphrase-robust or semantic watermarking schemes, where lexical divergence is not necessarily a good proxy for watermark disruption.

## Proposal

When both `--candidates N` (`N > 1`) and `--markllm-scheme` are enabled:

* Run the existing MarkLLM detector on every generated candidate.
* Include the following per-candidate measurements in `--json-stats`:
  * lexical divergence score
  * MarkLLM watermark score, when available
  * detector threshold, when available
  * `is_watermarked`
* Keep the current lexical candidate-selection behavior unchanged by default, preserving backwards compatibility.

For example:

```json
{
  "candidate_scores": [
    {
      "lexical_divergence": 0.91,
      "markllm_score": 4.3,
      "threshold": 3.0,
      "is_watermarked": true
    },
    {
      "lexical_divergence": 0.84,
      "markllm_score": 1.7,
      "threshold": 3.0,
      "is_watermarked": false
    }
  ]
}
```

This would make it possible to directly compare lexical divergence with watermark detection strength across rewrite candidates without changing the current selection policy.

## Alternatives

One alternative would be to add an opt-in detector-aware selector, for example:

```text
--candidate-selector lexical|markllm
```

which could select the candidate with the lowest watermark score.

However, I think exposing the per-candidate measurements first is a smaller and more conservative change. It preserves MarkLLM's current role as a verification harness rather than immediately turning the detector into an optimization oracle.

It would also provide useful evidence about whether lexical divergence is actually correlated with watermark removal before introducing detector-guided candidate selection.

## Extra context

The existing implementation already appears to contain most of the required pieces:

* candidate lexical divergence is already computed during multi-candidate selection;
* MarkLLM detection is already integrated into the Layer B rewrite workflow;
[...]

PR #110, opened 2026-08-17
fix: mark only one rewrite candidate as selected

ai-kunkun:
## Why

Per-candidate stats currently use `cand == out` to mark the selected rewrite candidate.

`_select_candidate()` selects a single candidate index, but if the rewrite backend returns duplicate candidate strings, multiple entries can compare equal to `out` and therefore be reported as `"selected": true`.

## What

- Track the selected candidate by index rather than string equality.
- Add a regression test covering duplicate rewrite candidates.
- Keep the existing lexical candidate-selection behavior unchanged.

## Verification

- `python3 -m pytest -q tests/test_rewrite_text.py`
- `python3 -m pytest -q`

Both test commands pass locally with Python 3.12. The full suite has one existing skip.

Related to #106 and #109.

### DIFF
--- a/service/scripts/rewrite_text.py
+++ b/service/scripts/rewrite_text.py
@@ -378,6 +378,7 @@ def rewrite(
     else:
         info["candidates"] = n
         out, scores = _select_candidate(text, outs)
+        selected_idx = max(range(len(outs)), key=lambda i: scores[i])
         trigger = markllm_scheme is not None or bool(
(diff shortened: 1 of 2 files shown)


poorvith-mp (approved):
Nice catch! Index-based tracking (`i == selected_idx`) cleanly prevents duplicate rewrite candidate strings from both being marked as `"selected": true`.

Tested against #109 and regression tests pass cleanly. Approved!

PR #111, opened 2026-08-17
fix: validate /clean option types

yearth:
## Summary

- validate `/clean` option values against the types declared in `ALLOWED_CLEAN_OPTIONS`
- return HTTP 400 when boolean or string options have the wrong JSON type
- add parameterized regression coverage for every supported clean option

## Problem

The OpenAPI schema declares options such as `nfkc` as booleans, but the
server only validates option names. A request containing `"nfkc": "false"`
is accepted and interpreted as true because `bool("false")` is `True`,
unexpectedly enabling NFKC normalization.

## Testing

- `python -m pytest -q`

### DIFF
--- a/service/scripts/server.py
+++ b/service/scripts/server.py
@@ -462,9 +462,13 @@ class Handler(BaseHTTPRequestHandler):
             options = {}
         if not isinstance(options, dict):
             raise ValueError("'options' must be an object")
-        for key in options:
+        for key, value in options.items():
             if key not in ALLOWED_CLEAN_OPTIONS:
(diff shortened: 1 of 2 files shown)


poorvith-mp (approved):
Thanks for catching and fixing this!

Validating `isinstance(value, expected_type)` against `ALLOWED_CLEAN_OPTIONS` prevents string truthiness bugs like `"false"` unexpectedly enabling flags like `nfkc`, and keeps the HTTP API strictly aligned with the OpenAPI spec.

Tests and CI checks look solid. Merging!

guillaumemeyer (approved):

PR #113, opened 2026-08-17
docs: add Windows auto-start guide via Task Scheduler 

Johar2503:
## What

Adds `docs/windows-autostart.md`: a guide for auto-starting `service/scripts/server.py` on Windows at login, using a silent `.vbs` launcher registered via Task Scheduler. No Docker required.

## Why

The README covers starting the service via Docker/compose or `make serve`, but neither persists across reboots/logins on Windows, and there's no documented path for Windows users without Docker. As a result, the `/remove-ai-marks` skill's health check (`WATERMARKS_SERVICE_URL`, default `http://127.0.0.1:8765`) fails until the service is started manually each session. This doc fixes that.

## Checklist

- Behaviour matches `SKILL.md` / `references/removal-matrix.md` when relevant—docs-only, no runtime behaviour changed
- Docs updated (README and/or skill references) if user-facing behavior changes
- No drive-by refactors unrelated to the fix or feature

## Notes for reviewer

Docs-only, no code touched. Scoped to local auto-start (`127.0.0.1` only)—deliberately does not cover exposing the service
over a public tunnel (e.g., ngrok), which is a different security posture and out of scope here. Manually verified on Windows 11: the service starts automatically at login with no visible console window, and `/health` responds correctly with zero manual steps. No sample files or secrets included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,10 @@ make serve                 # http://127.0.0.1:8765
 python3 service/scripts/server.py --host 127.0.0.1 --port 8765
 ```
 
+### Windows (no Docker)
+
+See [docs/windows-autostart.md](docs/windows-autostart.md) for auto-starting the service at Windows login without Docker.
+
 For the whole infra (core + optional harness/heavy backends), see [Docker / compose](#docker--compose) below.
 
 Optional system tools (auto-used when present — preinstalled in the core Docker image):
--- a/docs/windows-autostart.md
+++ b/docs/windows-autostart.md
@@ -1,94 +1,51 @@
-\# Windows: Auto-start the service at login
-
-
+# Windows: Auto-start the service at login
 
 Run the service silently in the background on every login, without a terminal window or manual start.
 
+## 1. Clone the repo somewhere permanent
 
+Replace `<path-to-clone>` below with wherever you want the repo to live (e.g. `C:\watermarks-remover` or `C:\Users\<you>\watermarks-remover`). Use the same path consistently in every step below.
 
-\## 1. Clone the repo somewhere permanent
+```powershell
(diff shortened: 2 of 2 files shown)


poorvith-mp (changes requested):
Thanks for putting together a Windows background startup guide! A few issues to fix before this can be merged:

1. **Markdown Formatting**: The file currently has literal backslash escapes on headers, code fences, and bullet points (`\#`, `\##`, code fences, `\-`) and extra newlines, which breaks markdown rendering.
2. **Paths**: `C:\watermarks-remover` is hardcoded across the commands—make sure to note that users should replace this with their actual clone path.
3. **Admin Elevation**: User-level logon triggers (`New-ScheduledTaskTrigger -AtLogOn`) don't require Administrator privileges—we should avoid asking users to elevate unnecessarily.
4. **Discovery**: Add a link to this doc in `README.md` under the server/deployment section so Windows users can find it.

Please clean these up and push an update!

Johar2503:
Thanks for the review! Pushed a fix addressing all four points:

1. Fixed markdown escaping—headers, code fences, and bullets are now clean (no more `\#`, `\##`, `\-`)
2. Replaced hardcoded `C:\watermarks-remover` with a `<path-to-clone>` placeholder, with a note to substitute their own path
3. Confirmed `-AtLogOn` doesn't need Administrator—removed the elevation instruction and added a note clarifying this
4. Added a link to the new doc under `### Start the service` in README.md

A new update has been pushed;  please review it.

poorvith-mp (approved):
Looks great now! Clean markdown formatting, clear path placeholders, no unnecessary admin elevation, and properly linked in README.md.

Approved!

PR #127, opened 2026-08-18
fix: detect AI markers in compressed PNG text chunks

yearth:
## Summary

- scan decoded `zTXt` and compressed `iTXt` text for existing AI/C2PA markers
- share the decoded marker results between PNG inspection and keep-metadata cleaning
- add regression coverage for both compressed PNG text formats

## Problem

PNG text metadata can be stored as compressed `zTXt` or compressed `iTXt`.
The parser already decompressed those chunks for generator-product matching,
but generic markers such as `Generated by` were still searched only in the
raw compressed payload.

As a result, inspection reported no AI metadata and
`--keep-non-ai-metadata` preserved an explicitly AI-tagged text chunk.

## Real-world reproduction

Pillow 12.3 generates both standard forms directly:

```python
from PIL.PngImagePlugin import PngInfo

ztext = PngInfo()
ztext.add_text("Comment", "Generated by AI", zip=True)

itext = PngInfo()
itext.add_itxt("Comment", "Generated by AI", zip=True)
```

Pillow reads `Comment = Generated by AI` back from both generated PNGs.

Before this change:

| Chunk | `has_ai_metadata` | Keep-metadata cleaning |
| --- | --- | --- |
| `zTXt` | `false` | chunk retained |
| compressed `iTXt` | `false` | chunk retained |

After this change, both variants report `has_ai_metadata: true` with a
`Generated by` finding, and keep-metadata cleaning drops the marked chunk.

## Testing

- `python -m pytest -q`
- Pillow 12.3 round-trip reproduction for `zTXt` and compressed `iTXt`


poorvith-mp:
Tested this against Pillow 12.3's `add_text(..., zip=True)` and `add_itxt(..., zip=True)` like your repro describes — confirmed the bug on `upstream/main` (`has_ai_metadata: false` on both, tagged chunk survives `--keep-non-ai-metadata`) and confirmed the fix on this branch (`true`, chunk dropped, `Generated by` finding).

Full suite: 435 passed, 6 skipped on this branch vs 433 passed, 6 skipped on main — the 2 new passes are your parametrized zTXt/iTXt cases, nothing else moved. Clean, single-commit, does one thing.

`_png_text_hits()` sharing the decoded-entries pass between the marker scan and the product-name scan is the right refactor — you're not decompressing twice, and `_text_chunk_is_ai` / `inspect_png` now actually agree on what "AI text" means for compressed chunks instead of one seeing raw bytes and the other seeing decoded text. Good catch.

guillaumemeyer (approved):

PR #93, opened 2026-08-16
fix: strip three missed Default_Ignorable invisible carriers in Layer A

PinoEire:
## Summary

Layer A's strip set tracks Unicode's `Default_Ignorable_Code_Point` property, but three **assigned, blank-rendering** carriers slipped through untouched, neither stripped by `clean_text` nor flagged by `inspect_text`, even between plain ASCII:

- **U+180F** MONGOLIAN FREE VARIATION SELECTOR FOUR (added in Unicode 14; siblings U+180B–U+180D were already handled)
- **U+3164** HANGUL FILLER (classic invisible / blank-name character)
- **U+FFA0** HALFWIDTH HANGUL FILLER

**Root cause:** their Unicode categories are `Mn`/`Lo`, so the `Cf` catch-all in `_decide` never saw them, and they were absent from `STRIP_CODEPOINTS`.

## Fix

- Add the three code points to `STRIP_CODEPOINTS`.
- Wire `U+180F` into the Mongolian-FVS handling so it is stripped when floating but **preserved after a Mongolian letter**, exactly like FVS1–3.
- Replace the `0x180B <= cp <= 0x180D` range with the named `_MONGOLIAN_FVS` set in the affected checks: clearer, and it avoids silently including the differently-handled Mongolian vowel separator at `U+180E`.
- Same fix applied to the vendored lightweight-skill engine (`skills/clean-user-facing-text/`).

The `inspect`-predicts-`clean` invariant is preserved: a full-range probe (all ~1.1M code points) shows zero under-reports both before and after.

## Tests

Three new cases in `tests/test_clean_text.py` (TDD: failing before, passing after):

- floating `U+180F` / `U+3164` / `U+FFA0` are stripped,
- `inspect_text` flags them,
- a regression guard that `U+180F` stays bound to its Mongolian base.

Full suite passes (`python -m pytest -q`); unchanged skips for absent optional torch backends.

## Scope note

The *unassigned* reserved ignorables (`U+2065`, `U+FFF0`–`U+FFF8`, `U+E0000`) are also passed through today. Stripping unassigned code points is a broader policy call rather than a missed-carrier bug, so it is intentionally left out of this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HV9AE8QLqiXf7QnRe891EY

PR #133, opened 2026-08-18
feat: consolidated Layer A hardening (missed carriers, reserved ignorables, noncharacters, layout format controls)

PinoEire:
Consolidates the four Layer A improvements from #93/#94/#95/#97 into a single branch rebased on current `main`, as requested in review. The four originals are closed in favour of this PR. The vendored-engine pin from #96 is already on `main` and is respected throughout: the vendored copy is byte-identical to the service engine in every commit.

## Contents (one commit per improvement, in original order)

1. **Strip three missed Default_Ignorable carriers**: `U+180F` (Mongolian FVS4), `U+3164` (Hangul filler), `U+FFA0` (halfwidth Hangul filler), blank-rendering `Mn`/`Lo` code points the `Cf` catch-all never saw.
2. **Strip reserved Default_Ignorable code points**: `U+2065`, `U+FFF0`-`U+FFF8`, `U+E0000`, `U+E0080`-`U+E00FF`, `U+E01F0`-`U+E0FFF`, reported under the new `reserved_ignorable` kind. Kept as explicit ranges rather than a category-Cn rule so a pinned `unicodedata` can never strip freshly assigned characters.
3. **Strip the 66 Unicode noncharacters** (`U+FDD0`-`U+FDEF`, `U+nFFFE`/`U+nFFFF`), reported under the new `noncharacter` kind. Permanently unassignable, so no future-Unicode risk.
4. **Preserve visible-layout format controls next to their own script**: Egyptian quadrat controls, Duployan overlap/step controls, musical beam/tie/slur/phrase controls; still stripped (and flagged) when floating between unrelated text.

## Review feedback addressed

- **U+3164/U+FFA0 vs U+115F/U+1160 inconsistency**: resolved by making the behaviour match the claim (a fifth commit). Both fillers are now preserved after a Hangul jamo of their own presentation form (compatibility jamo `U+3131`-`U+318E` for `U+3164`, halfwidth jamo `U+FFA1`-`U+FFDC` for `U+FFA0`), exactly like the conjoining fillers, and remain contraband when floating. The changelog states the rule precisely.
- **Stale description**: this PR body describes exactly what the branch contains.

## Testing

- `python -m pytest`: 458 passed, 3 skipped (optional-tool skips) on Windows.
- Exhaustive per-code-point tests for all 3,739 reserved ignorables and all 66 noncharacters, boundary tests proving assigned neighbours (`U+2064`, `U+FDF0`, `U+FFF9`, `U+FFFD`, `U+E0001`) are untouched, in-context preservation tests for every new preserve rule, and floating-strip tests for every new carrier.
- `ruff check` / `ruff format --check` clean on all touched files; the vendored engine parity test from #96 passes in every commit.


poorvith-mp:
Pulled this branch and ran the full suite locally: 455 passed, 6 skipped (your 458/3 split is the same result — the extra 3 skips on my end are just missing exiftool/qpdf/c2patool, not new gaps). Zero failures.

Walked the diff against the four originals it replaces. The fifth commit is the one I care about — it actually fixes the U+3164/U+FFA0 inconsistency instead of just rewording the changelog to match old behavior, which is the easy way out and you didn't take it. `_is_hangul_jamo` now covers compatibility (U+3131-U+318E) and halfwidth (U+FFA1-U+FFDC) jamo, so both fillers survive after their own presentation form and stay stripped when floating. That's the correct fix.

Reserved-ignorable ranges as explicit ranges instead of a Cn category rule is the right call too — pinning to a pinned `unicodedata` build means a Cn rule would eventually eat a freshly-assigned character. The comment telling future-you to re-check on Unicode version bumps is doing real work; U+180F becoming Mongolian FVS4 in Unicode 14 is exactly the failure mode it's guarding against.

Vendored/service parity holds in every commit, README table's updated. Nothing here I'd push back on.

guillaumemeyer:
@prvthmpcypher is that ready for merge in your opinion?

poorvith-mp:
@guillaumemeyer yes, in my opinion. Content-wise I already went through it, 455 passed and 6 skipped locally against the branch, walked the diff against the four PRs it replaces, and the Hangul filler fix (commit five) actually corrects the U+3164/U+FFA0 behavior instead of just rewording the changelog around it. Nothing there has changed since.

What's actually blocking it right now is git, not the code. I merged upstream/main into it locally to see exactly what's conflicting, and it's one spot: the Changelog section in README.md. Your branch and #153 (the iterative Layer B rewrite, merged since this was opened) both added an "### Unreleased" entry at the same anchor point, so git can't auto-merge two different insertions there. That's the entire conflict, nothing in the actual Layer A code collides with anything.

@PinoEire should be a two-minute fix on your end: rebase onto main, and where the changelog conflicts, keep both entries stacked (yours plus the Layer B one from #153), they're unrelated changes that both landed around the same time. Once that's resolved this should merge clean.

PinoEire:
@poorvith-mp rebased onto main, changelog entries stacked as suggested

poorvith-mp:
Rebase looks clean — checked the stacked changelog entries in \README.md\ and ran the full suite locally on the rebased branch: 542 passed, 6 skipped, 0 failed.

@guillaumemeyer good to merge from my side.

Issue #101, opened 2026-08-17
[feature] Multi-worker concurrency (-j) and SARIF 2.1.0 export for CI in audit_dir.py


poorvith-mp:
## Problem

`audit_dir.py` currently inspects directory trees sequentially on a single thread. When running over large codebases, documentation hubs, or media directories with hundreds of images and documents, audits take unnecessary time.

Additionally, when teams want to integrate directory audits into CI/CD pipelines (e.g. GitHub Actions), there is no standardized export format for surfacing findings directly in the GitHub repository's **Security -> Code Scanning** tab.

## Proposal

1. **Parallel Worker Threadpool (`--jobs` / `-j`)**:
   - Add concurrent scanning via `concurrent.futures.ThreadPoolExecutor` in `audit_dir.py`.
   - Default to `min(32, (os.cpu_count() or 1) + 4)` or user-configurable `-j N` (with `-j 1` for serial execution).
   - Keep results sorted by path for deterministic output across runs.

2. **OASIS SARIF 2.1.0 Export (`--format sarif` / `--sarif`)**:
   - Add a standardized SARIF 2.1.0 emitter in `audit_lib.py`.
   - Maps C2PA manifests, metadata markers, Layer A Unicode carriers, and stylometric signals into standard SARIF rules (`AI-WATERMARK-C2PA`, `AI-WATERMARK-METADATA`, `AI-WATERMARK-UNICODE-LAYER-A`, `AI-STYLES-HIGH-PROBABILITY`).
   - Enables direct ingestion by `github/codeql-action/upload-sarif@v3`.

3. **Format Flag**:
   - Support `--format text|json|sarif` (preserving `--json` and `--sarif` flags for ergonomics).

PR #138, opened 2026-08-18
feat: pre-commit hook integration (check + opt-in clean)

### LINKED ISSUE #135: Pre-commit hook integration for clean_file.py / audit_dir.py
CI catches AI-marked files before they merge — `audit_dir.py` got `-j` concurrency and SARIF export in #101 specifically so it could run there. But CI is the last checkpoint, not the first one. By the time it fires, the marked file's already committed, already pushed, maybe already reviewed. Fixing it at that point means an amended commit or a whole extra PR.

`git commit` is the natural place to catch this instead, and the project already has everything it needs to do it — this is a packaging problem, not a new-logic problem.

What I'd add: a `.pre-commit-hooks.yaml` at the repo root, exposing two hooks through the standard [pre-commit](https://pre-commit.com/) framework.

- `watermarks-remover-check` — runs staged files through `inspect_file.py`/`audit_dir.py`, fails the commit if it finds anything. Same check CI already runs, just earlier.
- `watermarks-remover-clean` — cleans staged files in place and re-stages them. I'd make this one opt-in specifically. Failing someone's commit is one thing; silently rewriting their staged diff is a bigger call, and that should be something a team chooses, not something that happens by default.

Once it's in, someone wires it up the normal pre-commit way:

```yaml
repos:
  - repo: https://github.com/guillaumemeyer/watermarks-remover
    rev: v0.5.0
    hooks:
      - id: watermarks-remover-check
```

No new detection logic in any of this — it just wraps `clean_file.py`/`audit_dir.py`, which already know how to walk a set of files and report or clean them. Stays stdlib-only too; a pre-commit hook entry just needs `language: python` and an `entry:` pointing at a script that already exists.

Only real open question is whether the auto-clean hook should even exist versus shipping just the check hook — rewriting someone's staged diff without asking is a bigger behavioral change than failing their commit with a message, so that one's the maintainer's call. Happy to send the PR.

poorvith-mp:
Closes #135.

CI already catches this — `audit_dir.py`'s SARIF export and `-j` concurrency from #101 exist specifically to run in CI. But that only fires after a marked file's already committed and pushed. This catches it earlier, at `git commit`, using the [pre-commit](https://pre-commit.com/) framework.

Added `.pre-commit-hooks.yaml` with two hooks:

- `watermarks-remover-check` — fails the commit and lists what it found when staged files carry AI/C2PA marks. Runs the same `scan_file()`/`is_actionable()` logic from `audit_lib.py` that `audit_dir.py` already uses for CI, so the pre-commit gate and the CI gate agree on what counts as actionable by construction — I didn't write a second definition of "actionable" anywhere.
- `watermarks-remover-clean` — opt-in, not the default. Cleans staged files in place, exits non-zero so you go review the diff and re-stage, same pattern `ruff --fix` and friends already use. Rewriting someone's staged diff without asking felt like a bigger call than just failing their commit with a message, so I didn't want that to be the thing that runs by default.

Both are thin wrappers — `check_staged.py` and `clean_staged.py` don't do any new detection or cleaning, they just call into what already exists. They also work standalone as plain CLIs if someone wants the same check outside the pre-commit framework.

`.pre-commit-hooks.yaml` needed a line added to the deny-by-default `.gitignore`, same as every other root config file already listed there.

## Testing

- `python -m pytest` — 442 passed, 6 skipped (main's at 433/6, the 9 new ones are `tests/test_precommit_hooks.py`)
- `ruff check` / `ruff format --check` clean
- Ran it by hand too: a file with an actual U+200B in it fails `check_staged.py`, `clean_staged.py` strips it and exits 1 so you know to re-stage, then `check_staged.py` on the cleaned file passes.

Added a "Pre-commit hook" section to the README with the config snippet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 !/.dockerignore
 !/.env.example
 !/.gitignore
+!/.pre-commit-hooks.yaml
 !/CODE_OF_CONDUCT.md
 !/compose-check.sh
 !/compose.yaml
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,16 @@
+- id: watermarks-remover-check
+  name: watermarks-remover check
+  description: Fail the commit if staged files carry AI/C2PA provenance marks.
+  entry: python3 service/scripts/check_staged.py
+  language: system
+  stages: [pre-commit]
+
+- id: watermarks-remover-clean
+  name: watermarks-remover clean (auto-fix)
+  description: >-
+    Strip AI/C2PA provenance marks from staged files in place. Opt-in: this
+    rewrites file contents, so add it only if you want staged files cleaned
+    automatically rather than just failing the commit.
+  entry: python3 service/scripts/clean_staged.py
+  language: system
+  stages: [pre-commit]
--- a/README.md
+++ b/README.md
@@ -780,6 +780,22 @@ Third-party projects that wrap or complement this repository, listed for discove
 
 To register a project here, open a PR adding a short entry — project name, what it wraps or adds, and a link to its own repository. Keep entries brief and factual; do not claim compatibility with, or endorsement by, this project. Please avoid names that start with or closely resemble `watermarks-remover` — look-alike names make it hard to tell which project is which.
 
+## Pre-commit hook
+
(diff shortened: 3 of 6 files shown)


guillaumemeyer (approved):

guillaumemeyer:
That's great @prvthmpcypher .
Since Claude Code is a watermarking code, you should advertise this feature more; share it on X and LinkedIn.

PR #140, opened 2026-08-18
fix: parse single-quoted attributes in relationship/manifest pruning

### LINKED ISSUE #130: Valid dangling relationships using single quotes in .rels files are incorrectly dropped, corrupting the OOXML container
### Description
In `container_meta.py`, `_prune_dangling_relationships` drops `<Relationship>` entries from `.rels` files if the target part no longer exists. However, the regular expression used to extract the `Target` attribute strictly expects double quotes: `re.search(r'\bTarget\s*=\s*"([^"]*)"', tag, re.I)`.

If a `.rels` file contains a perfectly valid XML relationship using single quotes (e.g., `<Relationship Target='word/document.xml'/>`), the regex fails to extract the target. As a result, `_target_attr` returns an empty string `""`, which resolves to the base directory. Because the base directory is not in `kept_names`, the script erroneously assumes the target was dropped and deletes the `<Relationship>` entry entirely.

This corrupts the OOXML file because critical relationships (like the main document part) are silently deleted.

### Environment
- OS: All
- Python version: All
- Component: `container_meta.py`

### Steps to Reproduce
1. Create a minimal DOCX file.
2. Edit `word/_rels/document.xml.rels` to use single quotes for a relationship: `<Relationship Target='document.xml' .../>`.
3. Run `python service/scripts/clean_file.py test.docx -o test_clean.docx`.
4. Observe the output: `- prune dangling relationships x1 in word/_rels/document.xml.rels`.
5. Try to open `test_clean.docx` in Microsoft Word (it will report the file as corrupt).

### Expected Behavior
The script should correctly parse the `Target` attribute regardless of whether it uses single or double quotes, and only prune the relationship if the target part was genuinely removed.

### Actual Behavior
Valid relationships using single quotes are unconditionally deleted, corrupting the archive.

### Root Cause
In `container_meta.py`, `_target_attr()` inside `_prune_dangling_relationships()` uses `r'\bTarget\s*=\s*"([^"]*)"'` which fails to match single-quoted attributes. (The same issue exists in `_prune_odt_manifest_entries` with the `full-path` attribute).

### Impact
DOCX, XLSX, and PPTX files generated by tools that use single quotes in XML attributes will be irreversibly corrupted upon cleaning.

### Severity
High

### Suggested Fix
Update the regular expressions in `_prune_dangling_relationships` and `_prune_odt_manifest_entries` to support both single and double quotes. For example: `r'\bTarget\s*=\s*(["\'])(.*?)\1'`.

yzxcj797:
## Summary

Fixes #130.

## Root cause

As reported: the `Target` extraction in `_prune_dangling_relationships` and the `full-path` extraction in `_prune_odt_manifest_entries` matched only double-quoted attribute values. A perfectly valid single-quoted `<Relationship Target='word/document.xml'/>` parsed as an **empty** target, which resolved to the package base directory, wasn't found in `kept_names`, and got the whole `<Relationship>` entry deleted — silently corrupting packages from tools that emit single quotes (Word then offers to repair).

## Fix

Both extractions now use a backreference pattern:

```python
re.search(r"\bTarget\s*=\s*([\"'])(.*?)\1", tag, re.I)
```

so either quote style resolves the real target, and pruning still only drops relationships/manifest entries whose parts were genuinely removed. `TargetMode` external handling and package-root cases are untouched.

## Tests

`tests/test_single_quoted_attributes.py`:

- a DOCX whose rels use single quotes for the kept main-document relationship survives cleaning intact (relationship present, **no** prune action reported) — this test fails on `main` with exactly the reported corruption;
- genuinely-dropped parts with double-quoted relationships still prune correctly (the fix doesn't loosen pruning).

Verified locally: new tests fail on `main`, pass with the fix; the full `tests/test_container_meta.py` suite stays green (43 passing combined).


poorvith-mp:
Confirmed, ran the suite: 435 passed, 6 skipped (main is 433, so the 2 new tests are the only difference, nothing else moved). The backreference pattern is exactly the right shape, matches whichever quote opened instead of assuming double. Applied to both `_target_attr` and `_full_path`, which is the part people usually forget about when fixing one and not the other. Good fix, nothing to add.

guillaumemeyer (dismissed):

guillaumemeyer (approved):
Re-approving after the lint-fix push (3a61246): CI lint + format + all checks green on the new head; same change as the previously approved version plus the unused-import/quote-style fixes.

PR #141, opened 2026-08-18
fix: report unmeasurable burstiness instead of scoring it as max LLM-like

### LINKED ISSUE #132: score_stylometry: empty sentence list scores as maximum LLM-likeness (fenced-body Markdown yields false positives)
## Summary

`score_stylometry.py` conflates "no sentences could be parsed" with "sentence lengths are perfectly uniform". Because `compute_burstiness()` returns `cv = 0.0` for an empty sentence list, and `cv < 0.25` is the *strongest* LLM signal tier (`burstiness_score = 0.95`), any file whose body text is skipped by the sentence extractor gets a **falsely elevated** score instead of being reported as unscoreable.

The most common trigger is a Markdown file whose entire body is wrapped in a fenced code block — but any unbalanced fence does it too.

## Reproduction

Same prose, only difference is the fence:

```bash
cat > plain.md <<'TXT'
We ran the probe for three weeks. Traffic was uneven.
Some days nothing arrived at all, and then a single autonomous system would saturate the link for
forty hours before disappearing without any obvious trigger, which made the early averages useless.
I assumed misconfiguration at first. It wasn't.
The second dataset disagreed with the first, which was inconvenient, because the whole analysis plan
had quietly assumed the two would move together. They do not move together.
One caveat remains. Coverage is thin.
TXT

{ echo '```tex'; cat plain.md; echo '```'; } > fenced.md

python3 service/scripts/score_stylometry.py plain.md  --json
python3 service/scripts/score_stylometry.py fenced.md --json
```

| file | score | level | `burstiness_cv` | `sentence_count` | `word_count` |
| --- | --- | --- | --- | --- | --- |
| `plain.md` | **0.0275** | CLEAN | 0.7258 | 11 | 82 |
| `fenced.md` | **0.3738** | LOW | 0.0 | **0** | 83 |

A 13.6x score increase on byte-identical prose.

## Expected

`sentence_count == 0` should yield an explicit "insufficient data" outcome (or at minimum drop the burstiness component from the composite), never the maximum LLM-likeness sub-score.

## Root cause

Two extractors disagree about whether fenced code is body text:

- `extract_sentences()` (line 153) skips fenced blocks, so it returns `[]`.
- `extract_words()` (line 175) runs on the raw text and does not, so `word_count` stays high — note `fenced.md` reports **83** words vs **82**, i.e. the literal `tex` from the fence line counts as prose.

The high `word_count` then defeats the small-sample dampener (`FULL_WEIGHT_WORDS = 100`, line 36), so on real-world chapter-length files the bogus score is not damped at all. On a ~1600-word file I saw a flat **0.4375** — exactly `0.95*0.45 + 0.0*0.45 + 0.1*0.10`, i.e. the entire score is the empty-sentence artifact.

`compute_burstiness()` (line 180) returns `0.0` for both "< 2 sentences" and a genuinely uniform text, so the caller at line 289 cannot distinguish them.

## Impact

Sentence extraction failing silently is the worst case for a detector meant for CI/CD gating (#68): the failure mode is a *false positive*, and it is invisible in the JSON output — `sentence_count: 0` is present but nothing flags it, and `notes` stays empty.

Real-world exposure is not marginal. Exporting a document library where each file stores its LaTeX/source inside a fence hit **159 of 162 files**, every one of them scored purely from this artifact.

## Suggested fix

1. In `compute_burstiness()`, signal absence explicitly (e.g. return `None` for `cv`, or a fourth `ok: bool`) rather than `0.0`.
2. At the scoring site, when burstiness is unavailable, either renormalize the composite over the remaining components or short-circuit to a `status`/`notes` entry such as `insufficient_data: no sentences parsed`.
3. Make `extract_words()` and `extract_sentences()` share one preprocessing pass, so `word_count` reflects the same text the sentence stats were computed from.

Happy to send a PR if the maintainers agree on which of (2)'s two behaviours is preferred.

## Environment

- commit `a2e72ed` (2026-08-17)
- Python 3.12.10, macOS 15 (arm64)
- `pytest`: 438 passed, 1 skipped before any change

yzxcj797:
## Summary

Fixes #132.

## Root cause (as reported, verified)

`compute_burstiness()` returned `cv = 0.0` both for "fewer than two sentences" and for genuinely uniform prose, and `cv < 0.25` is the **strongest** LLM-likeness tier (`0.95`, weight `0.45`). Any file whose body yielded no parseable sentences — most commonly a Markdown body entirely inside a code fence — therefore scored from the empty-sentence artifact alone (the issue's repro: **13.6x** score inflation on byte-identical prose; a real export hit **159/162** files), while `extract_words` still counted the fenced text so `word_count` defeated the small-sample dampener.

## Fix

- `compute_burstiness` returns **`None` for the CV** when unmeasurable (zero sentences, or fewer than two with words — a one-sample CV was equally reported as `0.0`).
- The scoring site **drops the burstiness component and renormalizes** the composite over the remaining components (`AI-phrase density 0.45` + `lexical diversity 0.10`, normalized by `0.55`), adding an explicit note: `Sentence burstiness unavailable (fewer than 2 parsed sentences — e.g. body wrapped in a code fence); composite renormalized over AI-phrase density and lexical diversity`.
- `burstiness_cv` is `null` in the JSON output for unmeasurable cases (including the `insufficient_length` early return, which also has no sentences), the human-readable CLI prints `n/a (fewer than 2 sentences)` instead of crashing on the `:.3f` format, and the uniform-cadence finding now guards on `cv is not None`.

## Tests

`tests/test_stylometry_empty_sentences.py` reproduces the issue's exact shape — byte-identical prose, plain vs fenced:

- the fenced variant keeps `sentence_count == 0`, reports a null CV plus the explanatory note, and scores in the same neighborhood as the plain file (the inflation collapses);
- plain prose keeps its real CV (no burstiness-unavailable note);
- `compute_burstiness` unit coverage for the `None` contract (empty and single-sentence).

Verified: the new tests **fail on `main`** with the reported inflation and pass with the fix; the existing stylometry suite (updated for the `None` contract) stays green.


poorvith-mp:
Confirmed, ran the suite: 436 passed, 6 skipped. Went with renormalize over short circuit for the composite when burstiness is unmeasurable, that's a reasonable call and matches one of the two options I flagged in the issue.

One thing worth knowing about, not blocking: word_count itself still counts the literal fence-line tokens, since extract_words doesn't skip fenced content the way extract_sentences does. The original repro's fenced.md reported 83 words against plain.md's 82. Doesn't matter for this bug since burstiness is fully out of the composite now when it's unmeasurable, but if word_count ever feeds something else downstream that gap is still sitting there. The actual bug, treating "unmeasurable" as "maximally uniform" and inflating the score 13.6x, is fixed.

guillaumemeyer (approved):

PR #142, opened 2026-08-18
fix: decode XML entities before Layer A scrub in OOXML/ODF

### LINKED ISSUE #129: XML entity encoded characters bypass Layer A scrub in OOXML/ODF containers
### Description
When `clean_file.py` processes OOXML (DOCX, XLSX, PPTX) and ODF (ODT) containers, it uses a regular expression to extract the text content and passes the inner XML string to `clean_text()`. However, `clean_text()` operates on the literal characters it receives and does not decode XML entities (e.g., `&#x200B;` or `&ZeroWidthSpace;`).

If a watermark generator injects zero-width characters into the document XML encoded as XML entities rather than literal UTF-8 bytes, `clean_text()` sees only the individual ASCII characters (e.g., `&`, `#`, `x`, `2`, `0`, `0`, `B`, `;`) and leaves them intact. When the file is re-assembled, the XML parser in Word/Writer resolves the entity back into the zero-width carrier, allowing the watermark to survive the cleaning process completely untouched.

### Environment
- OS: All
- Python version: All
- Component: `container_meta.py`

### Steps to Reproduce
1. Create a DOCX file where the XML body contains an entity-encoded zero-width space: `<w:t>Hello&#x200B;World</w:t>`.
2. Run `python service/scripts/clean_file.py test.docx -o test_clean.docx`.
3. Unzip `test_clean.docx` and inspect `word/document.xml`.
4. Observe that the `&#x200B;` entity remains completely unmodified.

### Expected Behavior
`clean_text()` (or the OOXML/ODF cleaner) should decode XML entities before applying the Unicode scrub, and re-encode necessary entities when writing back the XML, ensuring that invisible carriers cannot hide within XML entities.

### Actual Behavior
Entity-encoded characters bypass the Layer A scrub and survive the cleaning process.

### Root Cause
In `container_meta.py`, functions like `_scrub_docx_text` use `re.sub` and pass the raw `m.group(2)` to `clean_text(inner)`. Because `clean_text` operates at the character level and does not decode XML entities, encoded characters do not match the stripping lists.

### Impact
Watermarks embedded as XML entities completely bypass Layer A detection and removal.

### Severity
High

### Suggested Fix
Decode XML entities in the extracted text using `xml.sax.saxutils.unescape` (along with numeric entity decoding), pass the decoded string to `clean_text()`, and then re-escape it before substituting it back.

yzxcj797:
## Summary

Fixes #129.

## Root cause (as reported)

The OOXML/ODF scrubbers hand the raw text-run content to `clean_text()`, which works at the character level. A carrier encoded as `&#x200B;` arrives as **nine ASCII characters**, matches nothing in the invisible-character lists, and survives cleaning untouched — then the XML parser in Word/Writer decodes the entity back into the invisible carrier, watermark intact.

## Fix

A shared `_decode_xml_entities` resolves **exactly what a conforming XML parser resolves** — numeric character references (hex and decimal) and the five predefined entities — and leaves everything else literal (HTML5-only names like `&nbsp;`, which a strict XML parser would not resolve, and invalid code points like `&#0;`/`&#x110000;`). No new semantics are invented; the decoded string is what the reader will actually see.

- **DOCX/XLSX/PPTX** (`w:t` / `t` / `a:t` — text-only leaf elements): decode → `clean_text` → re-encode (`&`, `<`, `>`; a stable round trip for well-formed content).
- **ODT** (`text:p`): a whole-paragraph decode/re-encode would entity-escape the nested `text:span`/`text:tab`/`text:s` markup and corrupt the file — the scrubber now splits paragraph content into markup/text segments and only **text segments** are decoded, cleaned, and re-encoded; markup stays byte-identical.
- Runs where nothing was removed keep their **original bytes** (no entity churn for clean files).

## Tests

`tests/test_entity_encoded_watermarks.py`:

- the issue's exact repro: `Hello&#x200B;World` ends with the carrier removed in **both** forms (raw reference and decoded character);
- decode semantics: numeric + predefined resolved; `&nbsp;` and invalid code points stay literal;
- re-encode round-trip stability and ampersand (`Tom &amp; Jerry`) round-trip;
- clean text without carriers is byte-identical;
- the ODT case with nested span markup preserved while the entity-encoded carrier inside the same paragraph is removed.

The new tests fail on `main` (the helpers don't exist there, and the entity demonstrably survives cleaning); all touched suites pass with the fix, including the pre-existing ODT markup test that a naive whole-paragraph implementation would have broken.


poorvith-mp:
Confirmed, ran the suite: 439 passed, 6 skipped. This is more careful than what I sketched in the issue. I flagged over-escaping as a risk on the re-encode step and this handles it properly, only `&`, `<`, `>` get re-encoded, and HTML-only names like `&nbsp;` correctly stay literal since a real XML parser wouldn't resolve those either.

The ODT segmented markup handling is the part I'd have gotten wrong myself. A whole-paragraph decode and re-encode would have corrupted the nested `text:span` tags, splitting the paragraph into markup and text segments and only touching the text ones is the right call. Good fix.

guillaumemeyer (approved):

PR #143, opened 2026-08-18
fix: normalize script modes in the core image

### LINKED ISSUE #131: Core image fails to start when built with umask 027: scripts copied 0640, USER remover cannot read them
## Summary

`service/Dockerfile` copies `scripts/` without normalising file modes. On a build host with
`umask 027` the files land inside the image as `0640 root:root`, and since the image drops to
`USER remover` (uid 10001), **the process cannot read its own entrypoint**.

The image **builds successfully** and fails only at container start, which makes it easy to miss:
on a host with the more common `umask 022` the same Dockerfile produces `0644` and everything
works.

## Reproduction

```console
$ umask 027
$ git clone https://github.com/guillaumemeyer/watermarks-remover && cd watermarks-remover
$ podman build -f service/Dockerfile -t wr service/     # succeeds, no warning
$ podman run --rm --read-only --tmpfs /tmp wr
python3: can't open file '/app/scripts/server.py': [Errno 13] Permission denied
```

Confirming the mode inside the image:

```console
$ podman run --rm --entrypoint ls wr -ln /app/scripts/server.py
-rw-r----- 1 0 0 20828 /app/scripts/server.py
```

Tested at commit `07be21e54ab9`, podman 6.1.0, CachyOS (Arch), host `umask 027`.

## Cause

`Dockerfile:50` is `COPY scripts /app/scripts` and `Dockerfile:57` is `USER remover`.
`COPY` preserves the source file permissions, so the image inherits whatever umask the builder
happened to have. There is no `chmod` between the two.

## Suggested fix

One line, and it makes the build independent of the builder's environment:

```dockerfile
COPY --chmod=0755 scripts /app/scripts
```

Or, if you want to stay compatible with builders without BuildKit:

```dockerfile
COPY scripts /app/scripts
RUN chmod -R a+rX /app
```

The same consideration applies to the other `Dockerfile.*` variants if they copy scripts the same
way — I only exercised the core one.

## After the fix

With the permissions corrected the core service behaves exactly as documented, which is worth
recording since the hardening is good:

```console
$ podman run --rm -d -p 127.0.0.1:8765:8765 --read-only --tmpfs /tmp wr
$ podman exec <id> id
uid=10001(remover) gid=10001(remover) groups=10001(remover)
$ curl -s http://127.0.0.1:8765/health
{"ok": true, "version": "dev"}
```

And `/inspect` correctly reports `U+200B` and `U+00A0` on a crafted sample.

Thanks for pinning the base image by digest and `c2patool` by sha256 — that made the rest of the
review straightforward.

yzxcj797:
## Summary

Fixes #131.

## Root cause (as reported)

`service/Dockerfile` copies `scripts/` without normalizing modes, and `COPY` preserves the source file **and directory** permissions. On a build host with `umask 027` the scripts land as `0640 root:root` and `/app/scripts` as `0750` — unreadable and untraversable for the unprivileged runtime user (uid 10001). The image **builds successfully** and fails only at container start (`python3: can't open file '/app/scripts/server.py': [Errno 13] Permission denied`), which makes it easy to miss.

## Fix

Normalize after the copy with the canonical idiom:

```dockerfile
RUN chmod -R a+rX /app/scripts
```

- world-readable everywhere,
- execute bit for **directories** only (the `X` mode),
- intentionally-executable scripts keep their bit,
- deterministic regardless of the builder's umask — and independent of BuildKit `--chmod` directory-mode semantics.

The entrypoint is `python3 /app/scripts/server.py`, so the scripts need read, not execute.


poorvith-mp:
Confirmed, matches. Good catch on the directory mode too, I only flagged the file mode in the original issue but COPY preserves directory permissions the same way, so under umask 027 the directory itself lands at 0750 and even correctly-permissioned files inside would be untraversable. `chmod -R a+rX` handles both in one line, the execute bit only applies to directories and already-executable files so it's not blindly making everything executable.

guillaumemeyer (approved):

PR #145, opened 2026-08-18
feat: add reproducible SynthID-text removal benchmark

guillaumemeyer:
## What

Adds a reproducible, shareable **SynthID-text removal benchmark** — \`bench_synthid_text.py\` — that turns the repo's existing Layer B machinery (MarkLLM SynthID watermark generation/detection, \`rewrite_text.py\` rewrite hook, Layer A scrub) into a controlled experiment:

- Generates watermarked + unwatermarked corpora with the MarkLLM SynthID scheme (fixed seeds, same config for generation and detection, sanity gate per sample).
- Runs removal variants (strength × candidates) plus controls: no-removal baseline, Layer-A-only (expect ≈0% clear rate), and an optional re-stamp control that rewrites the unwatermarked text to detect backend re-stamping.
- Reports per variant: **clear rate**, **score suppression** (mean/median Δ), **quality** (lexical divergence, length drift, number/URL survival), and **cost** (estimated tokens in/out, wall time, optional USD) with a **clears-per-MTok** efficiency ratio.
- Optional **Gemini tier**: when \`WATERMARKS_GEMINI_API_KEY\` is set, every sample is also scored with Google's official SynthID-text detector.
- Emits \`report.md\` / \`results.json\` / \`results.csv\` per run; \`report.md\` embeds the exact reproduction command and pinned commits (repo + MarkLLM) so runs can be shared and rerun.

Also adds: bundled seed corpus under \`benchmarks/corpus/\`, \`docs/synthid-text-benchmark.md\` run/sharing guide, \`make bench-synthid-text\` target, and a \`.gitignore\` allow-rule for \`benchmarks/\`.

## Why

Layer B is documented as *best-effort*; nothing in the repo measured whether a given rewrite backend actually clears a SynthID-text-class mark, or at what cost. This gives users a way to (a) evaluate removal efficiency on their own model before trusting it, and (b) share a reproducible benchmark with peers instead of anecdotes.

## Checklist

- [ ] Behaviour matches \`SKILL.md\` / \`references/removal-matrix.md\` when relevant — N/A: no user-facing removal behaviour changes; this is a new optional tool
- [x] Unit tests updated or added under \`tests/\` — 14 mock-based tests, no torch/network in CI
- [x] \`python3 -m pytest -q\` passes (full suite green)
- [x] Docs updated — \`docs/synthid-text-benchmark.md\`, \`benchmarks/README.md\` (README changelog entry left for the release pass)
- [x] No drive-by refactors — only \`.gitignore\` (allow \`benchmarks/\`) and \`Makefile\` (new target)

## Notes for the reviewer

- **Two detection tiers, two honesty caveats** (both spelled out in the generated report): MarkLLM same-config detection is reproducible but is *not* Google's production SynthID-Text keying; the Gemini tier is the only vendor signal. Rewriting with a watermarked model can re-stamp — \`--restamp-control\` exists to check it.
- Heavy steps are external by design (matches the repo's optional-harness model): a MarkLLM checkout for watermark/detect, a real LLM backend for the rewrite. CI runs only the mock-based tests.
- The benchmark deliberately reuses \`detect_text_watermark.py\` / \`rewrite_text.py\` as subprocesses (the real product path), so it measures what actually ships.
- Statistical power: defaults are a quick smoke (\`--docs 3 --seeds 1\`); docs recommend ≥10 docs × ≥3 seeds.


guillaumemeyer:
@poorvith-mp you'll find this one interesting. Do not hesitate to suggest updates

poorvith-mp:
Went through this properly, not just the diff summary. Pulled the branch, ran the full suite (513 passed, 6 skipped, one unrelated flake in test_rewrite_text.py that passes clean on its own and on rerun, not caused by this PR), ruff clean, and read bench_synthid_text.py and the worker protocol end to end.

The methodology is the part that actually matters here and it's right. The sanity gate excludes any sample where the watermark wasn't even detected before removal, so a bad generation can't inflate the clear rate. The Layer A control is explicitly expected to fail, which is the correct control, if Layer A ever "cleared" a statistical mark that would mean the detector is broken, not that Layer A works. The weak-control note when the unwatermarked text itself flags positive is honest in a way a lot of benchmark scripts aren't.

MarkLLMWorker is careful in the specific ways that matter for a long-lived subprocess. The queue-plus-thread pattern in _read_line is the right way to put a timeout on a blocking pipe read, since readline() has none of its own. Draining stderr on a background thread avoids the classic deadlock where an unread pipe fills up and blocks the child. close() tries a graceful exit message before falling back to terminate, and test_run_cmd_no_rlimit_preexec locks in the no-RLIMIT_AS exception for MarkLLM explicitly, so nobody "fixes" that back in later and breaks CUDA without noticing why.

One real thing worth fixing: the PR description cites "#146 / commit 4f87ba0" as the source for the Gemini retirement. Commit 4f87ba0 is real, it's your own commit in this same branch's stack. But issue #146 is the DTD/entity-bomb regression test, completely unrelated. Same wrong citation shows up in #153's description too. Doesn't affect the code, just fix the reference before merge so nobody goes looking for context in the wrong issue.

On the removal itself: checked it's not a half-finished cleanup. vendor-notes.md cites the actual primary source (the Google AI forum thread confirming DETECT_TEXT_WATERMARK gets rejected on current models), and every remaining "gemini" hit in the codebase is either the unrelated image-metadata generator-name detection in container_meta.py or a legitimate mention of Gemini/SynthID as an image-pixel-watermark vendor class, which is a different feature entirely. Clean removal, well sourced, not scope creep.

Only nitpick I'd bother mentioning: median_score_delta uses sorted(deltas)[len(deltas) // 2], which is the lower element for an even-length list, not a true average of the two middle values. Doesn't change any conclusion at your default sample sizes, not blocking.

Good work. Approve.

PR #156, opened 2026-08-18
fix: treat a failed c2patool run as inconclusive, not as "no C2PA"

### LINKED ISSUE #155: [bug] a c2patool that fails to run is reported as "no C2PA", not as an inconclusive probe
## What happened

`run_optional_tools()` decides `has_manifest` purely by substring-matching c2patool's output. The exit code is recorded and never read. A c2patool that fails to run therefore produces `has_manifest: False` and nothing else, and `/inspect` answers `has_c2pa: false` — byte for byte the answer it gives for a genuinely clean asset.

`/capabilities` makes it harder to notice, because it reports `c2patool: true` from `which()`, which only proves the file exists.

I hit this on Apple Silicon. `service/Dockerfile` pins the base image by a digest that is a multi-arch index, so an arm64 host resolves it to `linux/arm64`, while the Dockerfile still installs the `x86_64-unknown-linux-gnu` c2patool release — upstream publishes no linux-aarch64 build. The binary dies before `main()`:

```
"returncode": -5,
"snippet": "rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2\n"
```

The arch mismatch is just the trigger I happened to have. The reporting hole is arch-independent: a kill, an OOM, a permission error, or a timeout all land in the same silent-negative path.

## What you expected

A probe that did not answer should be reported as inconclusive, not as a negative. `has_c2pa: false` should mean "checked, nothing found", never "the check crashed".

## Steps to reproduce

No arm64 host needed — the exit code and output are all that matter:

```python
import image_meta

def fake_run(cmd, **kw):
    class R: returncode = -5; stdout = ""; stderr = "rosetta error: failed to open elf\n"
    return R()

monkeypatch.setattr(image_meta, "which", lambda c: "/usr/local/bin/c2patool" if c == "c2patool" else None)
monkeypatch.setattr(image_meta.subprocess, "run", fake_run)

report = image_meta.inspect_image(Path("any.jpg"))
# report.has_c2pa is False
# report.notes  is []      <- nothing anywhere says the probe died
```

On an arm64 host, `docker compose up --build -d` then `GET /capabilities` reports `c2patool: true` while `docker run --rm --entrypoint c2patool <image> --version` fails.

## Environment

- OS and arch: macOS 15 (Darwin 25.6.0), arm64 / Apple Silicon
- Python version: 3.14 (in-image), 3.14.7 (host)
- How you run the skill: `skills/remove-ai-marks` against the compose service (`wr-core`, `docker compose up --build -d`)
- Optional tools: image built at `063119d`; c2patool 0.27.15 installed but not executable on arm64; exiftool 13.25 and qpdf 12.2.0 fine (Debian arm64 packages)

## Input type

- [x] Image (PNG / JPEG)
- [x] Both / batch directory

## Impact

Worst for PDF. `inspect_pdf()` notes that "exiftool/c2patool give more reliable metadata detection", and `container_meta.py` reads the same `run_optional_tools()` result, so a dead probe removes the detector the code itself calls the more reliable one.

For JPEG/PNG it costs corroboration rather than the only signal — the native APP11 and PNG-chunk scans still flag hard-bound C2PA. I verified that: with a c2patool that cannot run, a JPEG carrying a real signed manifest is still reported `has_c2pa: true` by the native scan.

## Extra context

Issues #1 and #3 fixed the opposite direction in this same function — `"No claim found"` matching `"claim"` and reporting a manifest on every clean asset. That fix added the negative-marker veto and `tests/test_c2patool_report.py`. This is the mirror case: the veto correctly suppresses the false positive, but nothing distinguishes "no manifest" from "no answer".

I have a fix with tests and will open a PR against this issue.

There is a second, separate question I did not touch: what the image should do about c2patool on arm64. Options seem to be forcing `--platform=linux/amd64` on the `FROM` (works, emulated, slow), selecting the asset by `TARGETARCH` and skipping c2patool on arm64 (honest once `/capabilities` is executability-based), or building it from source. That is a packaging call I would rather leave to you — with this fix in place, the skip option degrades safely instead of silently.

italovml:
## What

Makes a failed c2patool run reportable as its own state instead of collapsing into "no C2PA".

- `run_optional_tools()` now marks a run conclusive only when c2patool either found a manifest or said in so many words that there is none. A crash, a kill, or an unrecognized error sets `ok: False`. `has_manifest` keeps its current meaning and the existing negative-marker veto is untouched.
- `inspect_image()` and `inspect_pdf()` surface an inconclusive probe — as a `note` and a `finding` respectively. The wording avoids the substring `"c2patool reports"`, which `classify_finding_confidence()` maps to `confirmed`, and hits `"not fully inspected"` so it lands in `informational`.
- `/capabilities` version-probes each tool instead of trusting `which()`. Result is cached; a container's tool set cannot change while the process lives.
- One README line: `/capabilities` says "usable", not "present".

## Why

Fixes #155.

`run_optional_tools()` recorded `returncode` and never read it, so `has_manifest` came entirely from substring matching. A c2patool that died before `main()` produced `has_manifest: False` and no other signal, and `/inspect` answered `has_c2pa: false` — identical to the answer for a genuinely clean asset. `/capabilities` reported `c2patool: true` throughout, because `which()` only proves the file is on `PATH`.

Issues #1 and #3 fixed the false-positive direction of this same function. This closes the false-negative side.

Impact is largest for PDF, where `inspect_pdf()` itself documents c2patool as one of the more reliable detectors. For JPEG/PNG the native APP11 and PNG-chunk scans still catch hard-bound C2PA, so there it costs corroboration rather than the only signal.

## How I hit it

`service/Dockerfile` pins the base by a digest that is a multi-arch index, so an arm64 host builds an arm64 image and then installs the `x86_64-unknown-linux-gnu` c2patool release. It dies with `rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2`, exit `-5`.

Verified end to end on that arm64 image, before and after:

| | before | after |
| --- | --- | --- |
| `/capabilities` → `c2patool` | `true` | `false` |
| clean JPEG → `has_c2pa` | `false`, no note | `false` + inconclusive note |
| signed JPEG → `has_c2pa` | `true` (native APP11) | `true` + inconclusive note |

And on an amd64 build where c2patool genuinely runs, `/inspect` on a signed JPEG gives `returncode: 0`, `has_manifest: true`, `ok: true`, no note — so the honest path stays quiet.

## Checklist

## Notes for the reviewer

Layer: Files (C2PA/metadata). No fixtures added; the six new cases in `tests/test_c2patool_report.py` extend the existing `_fake_c2patool` helper, and one drives `server.capabilities()` directly.

`{"available": False}` for a missing c2patool is deliberately left byte-identical, so `test_missing_c2patool_is_reported_unavailable` still asserts exact equality.

Note that `run_optional_tools()` still calls `which()` rather than the new `_tool_usable()`: the probe should attempt the run and report its own outcome, while `/capabilities` answers the separate question of whether the tool is worth advertising. Happy to collapse them if you would rather have one path.

I deliberately left the packaging half alone — whether the image should force `--platform=linux/amd64`, select the c2patool asset by `TARGETARCH` and skip it on arm64, or build it from source. That is your call, and it is a different change. With this PR in, the skip option degrades honestly rather than silently.


poorvith-mp:
Pulled this and ran the full suite: 504 passed, 6 skipped, ruff clean. Nothing else moved.

This is the right fix and it goes further than the issue asked for. The core distinction, has_manifest versus ok versus conclusive, is exactly right: a crash or an unrecognized error now has has_manifest False and ok False, so nothing downstream can read a dead probe as a clean verdict. The c2patool_probe_note helper deliberately avoiding the substring c2patool reports so classify_finding_confidence doesn't bump it to confirmed is a detail I would have missed.

The /capabilities fix is the real bonus here. The issue only asked for the inspect-time reporting to stop lying, but which() proving a binary exists on PATH and not that it can run was a second, quieter version of the same bug: an arm64 host would report c2patool: true from /capabilities right up until you tried to actually use it. _tool_usable actually running --version and checking the exit code closes that gap too, and caching it per process is the correct call since a container's tool set can't change mid run.

Test coverage matches every path in the issue: the rosetta crash repro, an unrecognized error, both conclusive verdicts staying conclusive, and the exception path. Good fix.

guillaumemeyer:
@italovml detected a few items that may be improved:

Fragile coupling: the note's confidence bucket depends on magic substrings ("not fully inspected") in classify_finding_confidence(). Documented in the docstring, but a wording change silently reclassifies the finding.
Any chance to improve that?

ct.get("ok", True) in c2patool_probe_note silently assumes "ok" for dicts built by other paths: safe direction, but implicit.

poorvith-mp:
@italovml @guillaumemeyer both of these are real, and the first one has a clean fix already sitting in the codebase.

On the fragile coupling: classify_finding_confidence() already has an explicit escape hatch for exactly this situation, `t.startswith("info:")` is checked as its own condition ahead of the substring list, separate from incidentally-matching phrases like "not fully inspected." Right now c2patool_probe_note() lands in the informational bucket by coincidence of wording, not on purpose. Prefixing the returned string with `info:` (`f"info: c2patool probe inconclusive ({detail}); ..."`) makes the coupling explicit and load-bearing instead of implicit and one wording change away from silently reclassifying. One-word fix, and it's the pattern the function was already built to support.

On `ct.get("ok", True)`: agree it's implicit, though I'd call it low risk today specifically, `run_optional_tools()` is the only place that builds this dict and it sets `ok` explicitly on every branch, including the exception path. The gap is purely about some future caller building a c2patool tools-dict without knowing this default exists. Doesn't need a behavior change, a one-line comment above the `.get("ok", True)` stating "no ok key means an older/other code path built this dict; treat as conclusive rather than false-alarm" would make the implicit choice a documented one instead of a silent one.

Neither changes the actual logic I already tested, both are about making choices that are currently correct-by-coincidence into choices that are correct-on-purpose.

guillaumemeyer (approved):

PR #160, opened 2026-08-19
fix: install Archive::Zip so exiftool decodes Office containers

### LINKED ISSUE #154: [bug] core image lacks Archive::Zip, so exiftool cannot decode Office containers (docx/xlsx/pptx/odt/epub)
## What happened

The core image installs `libimage-exiftool-perl` but not the `Archive::Zip` Perl module. Without it, exiftool cannot decode ZIP-based containers, so for every Office/OpenDocument input it:

1. emits `Warning: Install Archive::Zip to decode compressed ZIP information`, and
2. **misidentifies the file as a generic ZIP** (`File Type: ZIP`, `MIME Type: application/zip`) instead of DOCX/XLSX/PPTX/ODT/EPUB, reading none of the embedded metadata.

The practical effect is that the exiftool arm of `/inspect` is blind for exactly the container formats the README advertises. The service's own DOCX/ODT parser still scans the parts, so results are not wrong — but the intended second, independent check silently degrades to nothing, and `report.tools.exiftool.interesting_lines` carries a warning instead of findings.

## What you expected

`Archive::Zip` present in the core image, so exiftool identifies the real container type and surfaces its metadata (e.g. `XMP-dc:Creator`) alongside the built-in parser.

## Steps to reproduce

Using a fixture from this repo and the published image:

```bash
docker run --rm --entrypoint exiftool \
  -v "$(pwd)/tests/fixtures:/data:ro" \
  ghcr.io/guillaumemeyer/watermarks-remover:latest /data/sample_ai.xlsx
```

## Environment

- OS and arch: container `linux/amd64`, Debian GNU/Linux 13 (trixie); host Windows 11 + Docker Desktop
- Python version: 3.14 (`python:3.14-slim` base, per `service/Dockerfile`)
- How you run the skill: `docker compose up -d` with the published `ghcr.io/guillaumemeyer/watermarks-remover:latest`; skill used as the documented HTTP client
- Optional tools present: `exiftool` 13.25, `c2patool`, `qpdf` — all reported `true` by `/capabilities`
- Perl 5.40.1; `perl -MArchive::Zip -e1` → `Can't locate Archive/Zip.pm in @INC`

## Input type

- [x] Both / batch directory

## Diagnostics

Stock image — note `File Type: ZIP`:

```
ExifTool Version Number         : 13.25
File Name                       : sample_ai.xlsx
Warning                         : Install Archive::Zip to decode compressed ZIP information
File Type                       : ZIP
File Type Extension             : zip
MIME Type                       : application/zip
```

Same image plus `libarchive-zip-perl` — correct type and readable contents:

```
ExifTool Version Number         : 13.25
File Name                       : sample_ai.xlsx
File Type                       : XLSX
File Type Extension             : xlsx
MIME Type                       : application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
Zip Required Version            : 20
```

On a real DOCX, `/inspect` goes from

```json
"interesting_lines": [
  "[ExifTool] ExifToolVersion : 13.25",
  "[ExifTool] Warning         : Install Archive::Zip to decode compressed ZIP information"
]
```

to

```json
"interesting_lines": [
  "[ExifTool] ExifToolVersion : 13.25",
  "[XMP-dc]   Creator         : Un-named"
]
```

which matches what the container's own `docProps/core.xml` holds.

## Suggested fix

One line in `service/Dockerfile`:

```diff
     && apt-get install -y --no-install-recommends \
         ca-certificates \
         curl \
+        libarchive-zip-perl \
         libimage-exiftool-perl \
         passwd \
         qpdf \
```

`libarchive-zip-perl` is 1.68-1 in trixie, pure Perl, no extra runtime deps. Verified locally as a thin `FROM ghcr.io/guillaumemeyer/watermarks-remover:latest` layer: read-only rootfs and the unprivileged `10001` runtime user are unaffected.

Worth considering as a follow-up: `/capabilities` currently reports `exiftool: true` in this state, which overstates what it can actually do for container formats. A capability probe for ZIP decoding (or surfacing the missing module) would make the degradation visible rather than silent.

## Extra context

Not a duplicate of #122 or #73, which touch DOCX handling in the scripts rather than the image's exiftool dependencies.

yzxcj797:
## Summary

Fixes #154.

## Root cause (per the issue's reproduction, verified against the Dockerfile)

The core image installs `libimage-exiftool-perl` but not its ZIP-container dependency `Archive::Zip`. Without it, exiftool on every Office/OpenDocument input:

1. emits `Warning: Install Archive::Zip to decode compressed ZIP information`, and
2. **misidentifies the file as a generic ZIP** (`File Type: ZIP`, `MIME Type: application/zip`) instead of DOCX/XLSX/PPTX/ODT/EPUB, reading none of the embedded metadata (no `XMP-dc:Creator`, etc.).

So the exiftool arm of `/inspect` — the intended second, independent check alongside the built-in container parser — is silently blind for exactly the formats the README advertises, and `report.tools.exiftool.interesting_lines` carries a warning instead of findings.

## Fix

One apt package: `libarchive-zip-perl` (Debian's packaged name for `Archive::Zip`) added to the same apt layer as `libimage-exiftool-perl`.

Verification against the issue's repro shape: `perl -MArchive::Zip -e1` will succeed in the rebuilt image, and `exiftool tests/fixtures/sample_ai.xlsx` will report the real `File Type` and embedded metadata instead of `application/zip` + the warning.


guillaumemeyer (approved):

poorvith-mp:
Confirmed, matches exactly what I verified on the issue. One package, correct name, correct spot in the layer alphabetically next to libimage-exiftool-perl. Ran the full suite on the branch too since I was already checking it out: 513 passed, 6 skipped, same as main, nothing else moved.

Only reason this isn't a bigger review is there isn't more to review, it's a one-line dependency fix that does exactly one thing. Good to merge.

PR #161, opened 2026-08-19
fix: distinguish a failed cleaner from an already-clean file in clean_staged.py (#159)

### LINKED ISSUE #159: [bug] clean_staged.py records a crashed clean_file.py as "skipped" and exits 0, so the commit proceeds uncleaned
## What happened

`_clean_one()` folds three different outcomes into one `"skipped"` bucket, and `main()` never lets a skip affect the exit code:

```python
# service/scripts/clean_staged.py:46-58
    if proc.returncode == 2:
        # Unrecognized format or oversized input; clean_file.py already
        # explained why on stderr, this is a non-fatal skip for the hook.
        return "skipped"
    if not proc.stdout.strip():
        eprint(f"{path}: {proc.stderr.strip() or 'clean_file.py produced no output'}")
        return "skipped"
    try:
        result = json.loads(proc.stdout)
    except json.JSONDecodeError:
        eprint(f"{path}: could not parse clean_file.py output")
        return "skipped"
```

The first branch is deliberate and I am not arguing with it: an unrecognized format is a real non-event, and `test_clean_staged_unknown_format_skipped` pins it. The other two are the problem. `proc.returncode` is never consulted again after the `== 2` check, so a `clean_file.py` that raised, was killed, or died before writing JSON lands in the same bucket as "nothing to do here". `main()` then returns 0 (`clean_staged.py:77-78`) because `changed_paths` is empty, and pre-commit shows a passing hook's output only in verbose mode, so the traceback on stderr is invisible in a normal run.

The hook's own docstring states the contract it breaks:

> this exits 1 whenever it modified at least one file, so the hook framework stops the commit and the developer reviews the diff and re-stages. Exit 0 means every file was already clean.

Exit 0 currently also means "the cleaner blew up on this file", which is the opposite conclusion.

## What you expected

A cleaner that failed to run should stop the commit, or at minimum be distinguishable from a file that was already clean. "Already clean" and "I could not clean it" cannot share an exit code on an auto-fix hook.

## Steps to reproduce

A staged symlink is enough. `safe_write_bytes()` refuses to write through one by design (`common.py:256`), which is correct, but `clean_file.py:115` calls `safe_write_text()` without a handler, so the refusal escapes as a traceback with an empty stdout:

```bash
mkdir -p /tmp/wm/real
printf 'Hello​World watermarked\n' > /tmp/wm/real/target.txt
ln -s /tmp/wm/real/target.txt /tmp/wm/link.txt

python3 service/scripts/clean_file.py /tmp/wm/link.txt --in-place --json; echo "exit=$?"
# Traceback (most recent call last):
#   ...
#   File ".../common.py", line 256, in safe_write_bytes
#     raise OSError(f"refusing to write through symlink: {dest}")
# OSError: refusing to write through symlink: /tmp/wm/link.txt
# exit=1

python3 service/scripts/clean_staged.py /tmp/wm/link.txt; echo "exit=$?"
# <same traceback, prefixed with the path>
# exit=0        <- commit proceeds
```

Afterwards the file is untouched and a sidecar is left behind:

```bash
python3 -c "print('U+200B present:', '​' in open('/tmp/wm/real/target.txt').read())"
# U+200B present: True

ls /tmp/wm
# link.txt  link.txt.bak  real
```

`backup_path()` runs at `clean_file.py:100`, before the write that fails, so every failed run also drops a `.bak` next to the file. This repo's `.gitignore` covers `*.bak`; a consumer repo installing the hook from `.pre-commit-hooks.yaml` will not.

The symlink is just the trigger that needs no special setup. Any unhandled error on the write path reaches the same place: a read-only directory, a full disk, `ENAMETOOLONG`, an unmounted network path, a file removed between staging and the hook run.

## Environment

- OS and arch: macOS (Darwin 25.5.0), arm64
- Commit: `063119d`
- How you run the skill: scripts only, invoked the same way `.pre-commit-hooks.yaml` invokes them
- Optional tools: none installed; not relevant, this is the stdlib text path

## Input type

- Layer involved: A (Unicode)

## Coverage

`tests/test_precommit_hooks.py` covers marked, already-clean, and unknown-format inputs. Neither the empty-stdout branch nor the `JSONDecodeError` branch has a test, which is consistent with them being unintended rather than designed.

## Related

Same shape as #158 (`scan_file()` swallows an `OSError` and the audit reports the file clean) and #155 (a c2patool that fails to run is recorded as "no C2PA"). Three sites, one pattern: a step that did not complete is recorded as a step that completed and found nothing. Worth deciding once whether "could not do the work" is allowed to share an exit code and a report shape with "did the work, nothing found" anywhere in this codebase.

codifly-network:
## What

Makes `clean_staged.py` distinguish "the cleaner could not run" from "the file was already clean". A run that produced no parsable report is now reported as a failure and exits `EXIT_PARTIAL` (3) instead of being folded into the `"skipped"` bucket and exiting 0.

- `_clean_one()` returns `(status, detail)` with a new `failed` status. The `returncode == 2` branch is unchanged — an unrecognized format is still a non-fatal skip that exits 0.
- `_failure_detail()` reduces a traceback to its last line (the actual cause), and names the exit status when a killed child left no stderr, instead of echoing the whole stack under a path prefix.
- `main()` reports failures in their own block, names the `.bak` sidecar left behind by the interrupted `--in-place` run, and returns `EXIT_PARTIAL if failures else (1 if changed_paths else 0)`.

## Why

Fixes #159. The hook's docstring promises "Exit 0 means every file was already clean"; today it also means the cleaner crashed. Because pre-commit only surfaces a passing hook's output in verbose mode, the traceback was invisible and the marked file went into the commit.

A staged symlink reaches it with no special setup: `safe_write_bytes()` refuses to write through one by design, `clean_file.py:115` calls `safe_write_text()` without a handler, so the refusal escapes as a traceback with empty stdout.

**The discriminator is deliberately not `returncode != 0`.** `clean_file.py` exits 1 with a valid JSON report for a *successful* clean that left residual signals, which `tests/test_json_exit_code.py` pins. Judging failure on the exit code would turn every residual-signal clean into a hook failure. The patch judges on whether a parsable report came back — no report means the run never reached its write.

**On exit 3:** 1 already means "your files were rewritten, review the diff and re-stage", and sending a developer to look for a diff that does not exist is the wrong instruction. 2 is usage/refusal, and is also the code `clean_file.py` returns for the skip that must still surface as 0. 3 is already defined in `common.py:19-23` as a partial run that outranks actionable findings, and `audit_dir.py:162` / `audit_website.py:555` already use it with the identical `return EXIT_PARTIAL if X else (1 if Y else 0)` shape. Happy to change this if you would rather the hook stayed binary.

The `.bak` is named rather than deleted: the wrapper cannot tell its own leftover from one the developer already had, and deleting the wrong file inside an already-failing hook is the worse outcome. `safe_write_bytes()` replaces atomically or not at all, so the original is intact and the sidecar is redundant. The ordering that creates it lives in `clean_file.py:100` and felt out of scope here.

## Checklist

## Notes for the reviewer

Layer: none directly — this is the pre-commit wrapper around `clean_file.py`.

8 new tests in `tests/test_precommit_hooks.py` cover the crash, killed-child, empty-stdout and malformed-JSON paths, plus three regressions that matter more than the new behaviour: the `rc == 2` skip still exits 0, a residual-signal clean (`rc == 1` with valid JSON) is still a success, and a failure outranks a successful clean in a mixed batch. The symlink case is end-to-end and needs no mocking; it uses the `_make_symlink` skip idiom from `tests/test_security_hardening.py` because CI includes `windows-latest`.

Verified in a `python:3.12-slim` container against the same commands CI runs: `pytest -q` 512 passed / 1 skipped (that skip is `test_pdf_structural_rewrite.py:145`, "needs real exiftool and qpdf", and is identical on `main`), `ruff check service tests` clean, `ruff format --check service tests` clean. I did not run the Windows or macOS legs.

No new dependencies, no changes outside `clean_staged.py`, its tests, and one README sentence.


poorvith-mp:
This is a real gap in the hook I wrote for #135, and the fix is right.

Confirmed the counts aren't a red flag first, since the branch is off an older main (missing #145 and #153), so raw pytest numbers don't line up against current main. Tested it properly against the actual merge-base instead: 498 passed/6 skipped there, 505 passed/7 skipped on this branch, so it's +7 tests passing and +1 skip (the symlink test correctly skipping on Windows CI). Tried to reproduce the symlink case by hand too, Git Bash without dev-mode privileges on Windows just silently falls back to copying instead of actually symlinking, which is exactly why the `_make_symlink` skip idiom exists rather than relying on a manual repro. So I'm trusting the 8 new tests here, and they cover the right things: the crash, the killed-child-with-no-stderr case, empty stdout, malformed JSON, and critically the two regressions that would've been easy to break by accident, `rc == 2` still exits 0, and a residual-signal clean (`rc == 1` with a valid report) still counts as success and not a failure.

The design call I care most about is judging on "did a parsable report come back" instead of on the exit code, and the reasoning holds up against the actual code: `clean_file.py` really does exit 1 for a successful clean that left residual signals, `test_json_exit_code.py` pins that, so treating any nonzero exit as a failure would have turned every residual-signal clean into a false hook failure. That would've been a worse bug than the one this fixes.

Exit 3 over 1 or 0 is the right call too. 1 already means "go look at a diff," and there's no diff to look at here, that's a worse instruction than no instruction. `EXIT_PARTIAL` already exists in common.py for exactly this "incomplete run" case and audit_dir.py already uses the identical `EXIT_PARTIAL if X else (1 if Y else 0)` shape, so this isn't a new convention, it's the hook actually following the one that was already there.

Naming the `.bak` instead of deleting it is the right defensive call too. The wrapper genuinely can't tell its own leftover from one the developer already had, and guessing wrong on a delete inside an already-failing hook is a strictly worse outcome than leaving a redundant file named in the output.

Good find, good fix. Nothing I'd change.

guillaumemeyer (approved):

PR #169, opened 2026-08-19
fix: an unreadable text file is a failed scan, not a clean one

### LINKED ISSUE #158: [bug] a text file that cannot be read is reported as clean, not as a failed scan
## What happened

`scan_file()` catches `OSError` on the text read and returns an item that carries an `error` key and nothing else:

```python
# service/scripts/audit_lib.py:48-51
    if kind == "text":
        try:
            text = path.read_text(encoding="utf-8", errors="surrogateescape")
        except OSError as e:
            return {"path": name, "kind": "text", "error": str(e)}
```

That item has no `confidence` key, so `is_actionable()` (`audit_lib.py:151`) reads `item.get("confidence", [])`, gets `[]`, and answers `False`. None of the three audit callers reads `item["error"]` either, so a file the scanner could not open is indistinguishable from a file it opened and found clean.

`classify()` returns early on a known extension without touching the file, so anything in `TEXT_EXTS` (`.txt .text .css .js .py .rs .go .json .yaml .yml .toml .csv`) reaches `read_text()` unguarded and the raise happens inside `scan_file`, where it is swallowed. Markdown and HTML are in `CONTAINER_EXTS` and take the `inspect_container()` route instead, where the `OSError` propagates normally, so this is the text branch only.

Three call sites inherit it:

- **`check_staged.py`** (the `watermarks-remover-check` pre-commit hook) calls `scan_file` directly, with no wrapper. An unreadable file produces no findings and the hook returns 0, so the commit proceeds.
- **`audit_dir.py`** wraps `scan_file` in `_scan_worker`, which correctly turns a *raised* exception into a `skipped` entry and `EXIT_PARTIAL`. But `scan_file` swallows the `OSError` before `_scan_worker` can see it, so the file is counted as scanned and the partial-scan signal never fires.
- **`audit_website.py:375`** has the same shape; least likely to trigger in practice, since it scans a temp file it just wrote.

This is the same class as #155 — a probe that did not answer being recorded as a negative answer — at a different site.

## What you expected

`actionable_files: 0` and exit 0 should mean "scanned, nothing found", never "could not read it". A file that failed to open belongs in `files_skipped` / `EXIT_PARTIAL`, which is what `common.py:20-23` already says the contract is:

> 3 = partial scan (some files/URLs failed to scan). A partial scan takes precedence over actionable findings: an incomplete audit is the more important CI signal.

## Steps to reproduce

No filesystem needed for the core of it:

```python
from audit_lib import is_actionable
is_actionable({"path": "notes.txt", "kind": "text", "error": "[Errno 13] Permission denied"})
# False
```

End to end, using the repo's own watermarked fixture:

```bash
mkdir -p /tmp/wm && cp tests/fixtures/sample_watermarked.txt /tmp/wm/secret.txt

# readable: the gate works
python3 service/scripts/check_staged.py /tmp/wm/secret.txt; echo "exit=$?"
# watermarks-remover: 1 file(s) carry AI/C2PA provenance marks:
#   /tmp/wm/secret.txt
#     - layer-a [zwj_family] U+200B ZERO WIDTH SPACE (Cf) x2
#     - layer-a [strip] U+00AD SOFT HYPHEN (Cf) x1
#     - layer-a [space] U+3000 IDEOGRAPHIC SPACE (Zs) x1
# exit=1

# same file, unreadable: silent pass
chmod 000 /tmp/wm/secret.txt
python3 service/scripts/check_staged.py /tmp/wm/secret.txt; echo "exit=$?"
# exit=0        <- no output at all
```

`audit_dir.py` over a directory holding only that one unreadable file:

```bash
python3 service/scripts/audit_dir.py /tmp/wm --json | \
  python3 -c "import json,sys; d=json.load(sys.stdin); print('files_scanned', d['files_scanned'], '| files_skipped', d['files_skipped'], '| actionable', d['summary']['actionable_files'])"
# files_scanned 1 | files_skipped [] | actionable 0

python3 service/scripts/audit_dir.py /tmp/wm --json >/dev/null; echo "exit=$?"
# exit=0        <- 3 (EXIT_PARTIAL) expected
```

The `error` string is present in the `--json` item and dropped everywhere else: the human report does not print it, SARIF emits no result for it, and the exit code ignores it.

## Environment

- OS and arch: macOS (Darwin 25.5.0), arm64
- Commit: `063119d`
- How you run the skill: scripts only, no service
- Optional tools: none installed (`c2patool` / `exiftool` / `qpdf` absent); not relevant to this path, which is pure stdlib text handling
- Caveat: the host had Python 3.9.6, below the documented 3.10+ floor, so `container_meta.py` needed `from __future__ import annotations` prepended to import at all (its `bytes | None` default annotation at line 152 is evaluated at def time). That is unrelated to this bug and the transformation does not touch `audit_lib`. The mechanism is version independent and readable straight from the source above.

## Input type

- [x] Both / batch directory
- Layer involved: A (Unicode)

## Impact

The two gates this breaks are the two places the tool is trusted to say no: the pre-commit hook and the CI audit. Both answer "clean" for content they never read, and CI sees a green build with an empty SARIF.

Permissions are the easiest trigger, but any `OSError` lands here — an I/O error, a dangling symlink, a file unlinked between `walk_files()` and the read, an unmounted network path mid-scan.

A fix in `_scan_worker` alone would not cover it: `check_staged.py` calls `scan_file` directly. The two shapes that do cover all three callers are re-raising so `_scan_worker` records the skip, or giving the error item a confidence that `is_actionable` treats as actionable.

yzxcj797:
## Summary

Fixes #158.

`scan_file`'s text branch catches the `OSError` and returns an item with an `error` key and **no `confidence` key** — `is_actionable` reads `item.get("confidence", [])`, gets `[]`, answers `False`. A file the scanner could not open was indistinguishable from one it opened and found clean:

- the pre-commit-style caller returned **0** (commit proceeds on an unscanned file);
- `audit_dir`'s `_scan_worker` never produced the `EXIT_PARTIAL` signal its try/except exists to produce, because `scan_file` swallowed the error **before** it could raise.

Same class as #155 — a probe that did not answer recorded as a negative answer.

## Fix — one deletion

The `except OSError: return {...}` block becomes `raise`. The wrappers already do exactly the right thing with a raised exception:

- `audit_dir._scan_worker` → `files_skipped` entry → `EXIT_PARTIAL`;
- `audit_website`'s wrapper → same raised-exception → skipped shape.

`common.py`'s documented contract ("a partial scan takes precedence over actionable findings: an incomplete audit is the more important CI signal") now actually fires.

## Tests

Regression test at all three layers: `scan_file` **raises** on an unreadable `.txt` (instead of returning a pseudo-clean item), the old error-item shape still reads non-actionable (guarding `is_actionable` semantics), and the worker produces the skipped entry with the OS error reason. Full `test_audit.py` (20) and `test_audit_sarif_and_concurrency.py` (5) stay green.


codifly-network:
Verified this closes #158. Letting the `OSError` propagate is the right shape — the wrappers that already exist do the correct thing with a raised exception, which is exactly what `scan_file` was preventing.

Measured as a non-root user (my first attempt ran as root in a container, where `chmod 000` is meaningless and the test silently passes for the wrong reason):

```
main    audit_dir exit=0  files_skipped=0  actionable=0     <- reported clean, never read the file
pr169   audit_dir exit=3  files_skipped=1  actionable=0     <- EXIT_PARTIAL
```

Two things worth raising before this merges.

## The branch is 17 commits behind, on a base predating the pre-commit hook

`git merge-base` puts this branch at `a2e72ed`, which is before `a3b4146` added `service/scripts/check_staged.py`. That file does not exist in this tree, so the pre-commit gate could not have been exercised here — and it is the caller with no wrapper around `scan_file`, so it is the one whose behaviour changes most.

Merged onto current `main` it does the right thing: the exception propagates out of `main()` and the hook exits non-zero, so the commit is blocked rather than silently passing. Full suite passes on the merge. Worth a rebase so CI tests what will actually land.

Minor UX point that follows from it: on the hook path the failure now surfaces as a raw traceback. It fails closed, which is the priority, but a caught `OSError` printing `cannot read <path>: <reason>` would read better than a stack trace in a pre-commit run.

## CI

```
service/scripts/audit_lib.py:49:27  F841  Local variable `e` is assigned to but never used
tests/test_audit.py:434:5           I001  Import block is un-sorted or un-formatted
```

`except OSError as e:` can drop the binding now that the body is just `raise`. Baseline `main` is lint-clean, so both are new.

Method: PR merged onto current `main`, run in `python:3.12-slim`, non-root for the permission cases.

guillaumemeyer:
Lint fixes pushed (222a0e2): the lint job was failing on two ruff findings — an unused `except OSError as e:` binding (F841) in `scan_file`, and an unsorted import block (I001) in the new regression test. Both are non-behavioral cleanups; the full local test suite (462 tests) and all CI checks pass.

guillaumemeyer (approved):

Issue #170, opened 2026-08-19
[bug] strip_png and strip_isobmff drop everything past an unparseable chunk and report "already clean", destroying recoverable images


codifly-network:
## What happened

`strip_png()` stops walking at the first chunk it cannot parse and never copies the remainder:

```python
# service/scripts/image_meta.py:1707-1708
        if chunk_end + 4 > len(data):
            break
```

Everything past that point is dropped from `out`. Then, because no chunk matched the removal list, the run is reported as a clean one:

```python
# service/scripts/image_meta.py:1742-1743
    if not actions:
        actions.append("no PNG metadata chunks removed (already clean or none matched)")
```

The result is written, `still_has_c2pa` / `still_has_ai_metadata` are false, `post_findings` is empty, and the exit code is 0.

`strip_isobmff()` has the same shape via the box walk at `image_meta.py:499-500`, and rebuilds only the boxes that parsed — so a truncated `mdat`, the actual coded image, is dropped.

## Why this is worse than a mangled file

The inputs are recoverable and the outputs are not. Built with Pillow, damaged the way an interrupted download damages a file, read back with `ImageFile.LOAD_TRUNCATED_IMAGES = True`:

```
truncated mid-IDAT       in=89B  out=33B  exit=0  actions=['no PNG metadata chunks removed (already clean or none matched)']
                         INPUT : RECOVERABLE (64, 64)
                         OUTPUT: UNRECOVERABLE UnidentifiedImageError
```

A real AVIF (`pillow-avif-plugin`, 319 bytes) truncated to 223:

```
in=223B  out=32B  exit=0  actions=['no AVIF metadata boxes removed (already clean or none matched)']
```

86% of the file dropped, reported as an already-clean image.

So this is not "the cleaner mangles a corrupt file". A user who still had a viewable photo has an unopenable husk afterwards, and the tool said it was fine.

## The sibling formats already do this correctly

Same 60%-truncation damage, same command:

| format | result |
| --- | --- |
| **WebP** | exit **1** — `error: malformed WebP: RIFF size mismatch: header=94 actual=56; truncated WebP chunk VP8` |
| JPEG | preserved, and named: `"truncated segment; copied remainder"` (`image_meta.py:1791`) |
| BMP | preserved, and named: `"BMP header not fully parsed; left unchanged"` (`image_meta.py:666`) |
| GIF | preserved — copies the remainder (`image_meta.py:831`, `859`) |
[...]

Issue #167, opened 2026-08-19
[bug] a truncated AVIF/HEIC/MP4 skips the whole-file C2PA byte-scan its sibling inspectors run


codifly-network:
## What happened

`inspect_isobmff()` returns early when no boxes parse, jumping over the whole-file C2PA byte scan that every sibling inspector reaches:

```python
# service/scripts/image_meta.py:512-514
    boxes = _parse_isobmff_boxes(data)
    if not boxes:
        return False, False, [f"not a valid {fmt.upper()} (no ISOBMFF boxes found)"]
```

The fallback it skips is at `:571-574`:

```python
    whole = _contains_any(data, C2PA_MARKERS)
    if whole and not has_c2pa:
        has_c2pa = True
        findings.append(f"byte-scan C2PA markers: {', '.join(whole[:6])}")
```

`inspect_png:355-359`, `inspect_jpeg:411-414`, `inspect_gif:789-792` and `inspect_tiff:1124-1127` all reach their equivalent even after a truncation.

## Steps to reproduce

One AVIF, two states, differing only in the first box's declared size:

```
intact                   -> has_c2pa=True   findings=['byte-scan C2PA markers: c2pa, C2PA, jumb, JUMB, contentcredentials']
first box size overruns  -> has_c2pa=False  findings=['not a valid AVIF (no ISOBMFF boxes found)']
                            literal b'c2pa' present in the file: True   (in both cases)
```

A first box whose size field overruns the file is what a truncated AVIF/HEIC download produces: `_parse_isobmff_boxes:499-500` breaks and returns `[]`.

Also reached for MP4/MOV/M4A through `av_meta._inspect_mp4:132`.

## What you expected

The byte-scan fallback exists for exactly this case. A container the box parser could not walk is where it is most useful, not least.

`"not a valid ..."` maps to `informational` in `classify_finding_confidence` (`common.py:377`) and so never moves an exit code.

## Environment

- macOS (Darwin 25.5.0), arm64; commit `063119d`

## Input type

- [x] Image (PNG / JPEG)
- Layer involved: Files (C2PA/metadata)

yzxcj797:
Fixed — the early return now runs the whole-file C2PA byte scan before reporting the parse failure, mirroring what every sibling inspector does after a truncation:

- a walkable container keeps the box-level findings (the stronger evidence — a named `c2pa`/`jumb` top-level box);
[...]

Issue #164, opened 2026-08-19
[bug] one unreadable zip member discards the C2PA/AI markers already found in earlier members


codifly-network:
## What happened

The container inspectors accumulate `has_c2pa` / `has_ai` / `findings` member by member, then discard all of it if any later member fails to read:

```python
# service/scripts/container_meta.py:884-885 (OOXML), :1351-1352 (ODT), :1534-1535 (EPUB)
    except _ZIP_PARSE_ERRORS:
        return False, False, [f"not a valid {fmt.upper()} zip"], {}
```

Markers already found in earlier members are thrown away and replaced with hardcoded `False, False`.

## What you expected

Evidence already collected should survive a later read failure, and a container that could not be fully read should not be reported as one that was read and found clean.

The repo already holds this principle elsewhere: `tests/test_container_meta.py:783-793` pins that a refused zip bomb must propagate rather than "be reported as an unparseable container". The existing malformed-zip test at `:772-780` only covers wholly-garbage bytes, where returning `False` is correct. The partial-read case is uncovered.

## Steps to reproduce

A DOCX whose `docProps/core.xml` carries `OpenAI ChatGPT` and `c2pa contentcredentials`, plus one `customXml/item1.xml` whose compressed payload is corrupted while the central directory stays intact:

```
intact           has_c2pa=True   has_ai=True   findings=['docProps/core.xml: marker:c2pa, marker:C2PA, marker:contentcredentials, ai:ContentCredentials, ai:OpenAI', 'customXml parts: 1']
one bad member   has_c2pa=False  has_ai=False  findings=['not a valid DOCX zip']
```

(the bad member raises `zlib.error: Error -3 while decompressing data: invalid code lengths set`)

At the gate:

```
$ python3 service/scripts/check_staged.py intact.docx  ; echo $?   -> 1
$ python3 service/scripts/check_staged.py corrupt.docx ; echo $?   -> 0
```

Real-world triggers, all in `_ZIP_PARSE_ERRORS` (`container_meta.py:65-74`): a corrupt deflate stream (`zlib.error`), a CRC mismatch (`BadZipFile`), a compression method `zipfile` cannot decode (`NotImplementedError`), an encrypted member (`RuntimeError`).

[...]

PR #171, opened 2026-08-19
fix: report and drop truncated ID3v2 tags instead of certifying them

### LINKED ISSUE #163: [bug] a truncated ID3v2 tag reports zero findings, and clean_file.py then certifies the file clean
## What happened

`_parse_id3v2_frames()` gives up on a tag whose declared size overruns the file, and `_inspect_id3v2()` converts that into a clean result:

```python
# service/scripts/av_meta.py:173-174
    total = 10 + tag_size
    if total > len(data):
        return None

# service/scripts/av_meta.py:200-201
    parsed = _parse_id3v2_frames(data)
    if parsed is None:
        return False, False, []
```

`detect_av_format()` still returns `"mp3"`, so the file is claimed as inspected. The report is not merely empty of findings — `notes` is empty too, so it is byte-for-byte what a genuinely clean MP3 produces.

`_strip_id3v2()` at `:228-229` takes the same branch, so the cleaner writes the file back verbatim and the post-clean verification confirms the result.

## What you expected

A tag the parser could not read should be reported as unread. `has_c2pa: false` with no notes should mean "parsed the tag, nothing there".

## Steps to reproduce

An MP3 with four `TCOM` frames carrying `c2pa` / `contentcredentials` / `OpenAI ChatGPT` / `Generated by AI`, then the same file truncated:

```
full.mp3  (173 B)  has_c2pa=True   has_ai=True   findings=['ID3v2 frame TCOM: c2pa, C2PA', 'ID3v2 frame TCOM: contentcredentials, ContentCredentials', 'ID3v2 frame TCOM: OpenAI', 'ID3v2 frame TCOM: Generated by']
cut.mp3   ( 86 B)  has_c2pa=False  has_ai=False  findings=[]  notes=[]
```

Then the cleaner on the truncated one:

```
$ python3 service/scripts/clean_file.py cut.mp3 -o cut.cleaned.mp3 --json
{"format": "mp3", "actions": [], "bytes_in": 86, "bytes_out": 86,
 "still_has_c2pa": false, "still_has_ai_metadata": false, "post_findings": []}
$ echo $?
0
$ python3 -c "print(b'contentcredentials' in open('cut.cleaned.mp3','rb').read())"
True
```

The marker is in the bytes the cleaner just wrote out and reported clean.

`check_staged.py` exits 0 on it, and `.pre-commit-hooks.yaml` sets no `types:` filter, so `.mp3` genuinely reaches the hook.

## What is different about the image path

`inspect_png:329-331` and `inspect_jpeg:391-393` both record a truncation finding and still reach the whole-file byte scan. The AV path has neither.

## Environment

- macOS (Darwin 25.5.0), arm64; commit `063119d`; scripts only, no optional tools (not relevant to this path)

## Input type

- [x] Both / batch directory

## Related

Same class as #158, #159 and #155: a step that did not complete recorded as a step that completed and found nothing. This one is the worst of the set, because the cleaner does not just misreport — it hands back a file it has certified.

yzxcj797:
## Summary

Fixes #163.

## Root cause (per the issue's reproduction, verified)

`_parse_id3v2_frames` returns `None` when the tag's declared size overruns the file, and **both** consumers converted that into a clean result:

- `_inspect_id3v2` → `return False, False, []` — empty findings **and** empty notes, byte-for-byte what a genuinely clean MP3 produces (the issue's `cut.mp3`: markers present in the readable prefix, `has_c2pa=False, notes=[]`);
- `_strip_id3v2` → `return data, []` — the cleaner writes the file back **verbatim**, the post-clean verification confirms it, `clean_file.py` exits 0. A truncated MP3 is certified clean end to end.

## Fix — both halves

1. **Inspect**: when the `ID3` header itself parsed but the declared size overruns, return a finding naming the truncation and stating the negative means *unread*, not *clean*. A file with no `ID3` header at all keeps the plain empty result (nothing to misreport).
2. **Strip**: drop the readable prefix of a truncated tag **unconditionally** — frame boundaries are unknowable, so a whole-tag drop is the only safe clean, and it removes whatever markers the readable prefix carried. The action records the truncation: `drop truncated ID3v2 tag (N bytes of M declared)`.

## Tests

`tests/test_id3_truncated.py`, using the issue's exact shape (declared 170-byte tag, 68 bytes available, `contentcredentials` in the readable prefix):

- inspect surfaces the truncation finding;
- a well-formed small tag parses without the note;
- strip removes the tag and the marker, recording the truncated-drop action.

**Both discriminating tests fail on current `main`** (inspect returns the byte-for-byte-clean result; strip is a verbatim no-op). Full `test_av_meta.py` stays green alongside.

## Summary by CodeRabbit

- **Bug Fixes**
  - Truncated ID3v2 metadata is now correctly identified as unreadable instead of being reported as clean.
  - Cleaning truncated metadata now removes the incomplete tag and any readable remnants.
  - Complete, valid ID3v2 tags continue to be processed normally.
  - Metadata cleanup no longer preserves damaged tag data.

- **Tests**
  - Added coverage for detecting truncated tags, safely removing incomplete metadata, and continuing to process valid tags correctly.


aadarshbk (approved):

codifly-network:
Thanks for taking this one. The inspect half is right — the finding names the truncation and says plainly that the negative means unread. Two problems on the clean half, and the first is a blocker.

## The whole-tag drop deletes the whole file

```python
tag_size = _id3v2_size(data, 6)
total = min(10 + tag_size, len(data))
return data[total:], [...]
```

The parser only reached this branch because `10 + tag_size > len(data)`, so `min(...)` always collapses to `len(data)` and `data[total:]` is always `b""`. Every file that takes this path is emptied.

The damaging case is not a truncated download — it is a **complete, valid file with one corrupted byte in the syncsafe size field**. A 1063-byte MP3 (35-byte tag + 1028 bytes of audio), one byte flipped at offset 6, run against this branch on a merge of this PR onto current `main`:

```
in=1063B  out=0B  exit=0
actions=['drop truncated ID3v2 tag (1063 bytes of 266338339 declared)']
audio bytes survived? False
```

All 1028 bytes of audio are gone, and the action line reports a tag drop. On `main` today that same file comes out intact at 1063 bytes. So this trades "certifies a marked file clean" for "silently deletes the file", which is the shape of #170.

`if len(data) >= 10 and data[:3] == b"ID3"` also does not sanity-check `tag_size` — a single flipped bit turns 35 into 266,338,339.

Suggestion: refuse rather than guess, matching the WebP idiom the repo already blesses in `test_truncated_webp_is_reported_and_not_rewritten` — `raise ValueError("malformed MP3: ID3v2 tag is truncated ...")`, which `clean_file.py:137` already turns into exit 1. If you would rather still clean, resynchronising to the next MPEG frame header and dropping only up to there preserves the audio; a blind whole-file drop cannot.

A test asserting the audio survives would catch this — the current suite passes with the file emptied.

## The gate still passes

The new finding is honest, but `classify_finding_confidence` matches none of its substrings and falls through to `informational`, so `is_actionable` stays False:

```
check_staged.py on a truncated-tag mp3 carrying c2pa/OpenAI markers -> exit 0, silent
confidence mapping: ['informational']
is_actionable: False
```

That is the half of #163 about the gate. Worth deciding whether "could not read this" should be actionable — it is the same question behind #164 and #167.

## CI

`ruff format --check` fails at `service/scripts/av_meta.py:206`, and `tests/test_id3_truncated.py:18` has an unused `# noqa: E402` (`ruff.toml` already grants `tests/**` that ignore). Baseline `main` is lint-clean.

codifly-network (changes requested):
Converting my earlier comment into a formal review, because this PR currently shows an approval and the blocking problem is data loss rather than a style question.

**As written, this branch empties files.**

```python
tag_size = _id3v2_size(data, 6)
total = min(10 + tag_size, len(data))
return data[total:], [...]
```

The parser only reaches this branch because `10 + tag_size > len(data)`, so `min(...)` always resolves to `len(data)` and `data[total:]` is always `b""`. Every file that takes this path is emptied — not only truncated downloads.

Measured on a merge of this branch onto current `main`, in `python:3.12-slim`, on a **complete, valid 1063-byte MP3** (35-byte tag + 1028 bytes of audio) with a single flipped byte at offset 6:

```
in=1063B  out=0B  exit=0
actions=['drop truncated ID3v2 tag (1063 bytes of 266338339 declared)']
audio bytes survived? False
```

On `main` today that same file comes out intact at 1063 bytes. So the change trades "certifies a marked file clean" for "silently deletes the file", which is the failure shape of #170. The action line reports a tag drop while every audio frame is gone, and `if len(data) >= 10 and data[:3] == b"ID3"` does not sanity-check `tag_size` — one flipped bit turns 35 into 266,338,339.

The inspect half is good and I would keep it: the finding names the truncation and states that the negative means unread.

Two smaller points from the full comment above, neither blocking on its own: the new finding maps to `informational` in `classify_finding_confidence`, so `check_staged.py` still exits 0 on a marked file; and CI `lint` is failing (`RUF100` in the new test plus `ruff format` at `service/scripts/av_meta.py:206`).

Suggested direction, using the idiom this repo already blesses in `test_truncated_webp_is_reported_and_not_rewritten`: `raise ValueError("malformed MP3: ID3v2 tag is truncated ...")`, which `clean_file.py:137` already turns into exit 1. If you would rather still clean, resynchronising to the next MPEG frame header and dropping only up to there preserves the audio — a blind whole-file drop cannot.

A test asserting the audio survives would catch this; the current suite passes with the file emptied.

guillaumemeyer:
Fixing the linting issue rn

guillaumemeyer:
@codifly-network any progress on this? Need help? 
image

guillaumemeyer:
@coderabbitai full review

guillaumemeyer:
@coderabbitai full review

poorvith-mp (changes requested):
The inspection finding is helpful, but the stripping implementation causes total data loss on truncated or corrupted headers:

```python
tag_size = _id3v2_size(data, 6)
total = min(10 + tag_size, len(data))
return data[total:], [...]
```

When `10 + tag_size > len(data)`, `min(...)` always evaluates to `len(data)` and `data[total:]` returns `b""`, wiping the entire audio file down to 0 bytes.

To fix this safely without data loss:
1. Don't drop `len(data)` blindly. Scan forward for the next valid MPEG audio frame sync (`0xFF` with valid MPEG audio header bits) and preserve audio from that sync offset.
2. If no valid frame sync exists, raise `ValueError("malformed MP3: truncated ID3 tag with no valid audio frames")` so `clean_file.py` exits 1 without overwriting the target file with 0 bytes.
3. Add a test asserting that audio payload bytes are retained after cleaning a truncated-tag MP3.

bentalay:
@yzxcj797 # Fix plan — PR #171

1. **Stop discarding recoverable audio on truncation.** — `service/scripts/av_meta.py:276-278` — Before treating a parse failure as truncation, verify the declared size actually overruns the file; when dropping the tag, scan the remainder for a valid MPEG frame sync (footer-aware for v2.4) and preserve that suffix, otherwise return the input unchanged with a "refused unsafe clean" action instead of emptying the output. — **L**

2. **Don't drop complete-but-malformed tags via the truncation branch.** — `service/scripts/av_meta.py:276-278` — The branch fires on any ID3-prefixed parse failure, including full tags with bad frames (non-target frames included, even with `strip_all_metadata=False`); gate it on `10 + tag_size > len(data)` and route non-overrun failures to the refuse path from item 1. — **S**

3. **Make a failed probe inconclusive, not clean.** — `service/scripts/av_meta.py:229-241` — `has_c2pa`/`has_ai` still read `False` on an unreadable tag; return an explicit tri-state (e.g. `None`/`"inconclusive"`) or a status field and propagate it through report output and exit aggregation so callers can't certify. — **M**

4. **Add payload-preservation tests.** — `tests/test_id3_truncated.py:49-56` — Add cases asserting byte-identical audio tail after cleaning: a truncated tag followed by a valid MPEG frame, a complete-but-unparseable tag with trailing payload, and an ID3v2.4 tag with footer. — **M**

5. **Add a refusal/inconclusive regression test.** — `tests/test_id3_truncated.py` — Assert the non-overrun malformed-tag path returns input unchanged with a refusal action, and that inspection reports the inconclusive status from item 3. — **S**

service/scripts/av_meta.py
```diff
@@ -254,6 +268,16 @@ def _inspect_id3v2(data: bytes) -> tuple[bool, bool, list[str]]:
 def _strip_id3v2(data: bytes, *, strip_all_metadata: bool) -> tuple[bytes, list[str]]:
     parsed = _parse_id3v2_frames(data)
     if parsed is None:
+        if len(data) >= 10 and data[:3] == b"ID3":
+            # Truncated tag: frame boundaries are unknowable, so a whole-tag
+            # drop is the only safe clean. Dropping it unconditionally also
+            # removes whatever markers the readable prefix carried, instead of
+            # writing the file back verbatim and certifying it clean (#163).
+            tag_size = _id3v2_size(data, 6)
```
bentalay, line 282: Stop discarding recoverable audio on truncation. — service/scripts/av_meta.py:276-278 — Before treating a parse failure as truncation, verify the declared size actually overruns the file; when dropping the tag, scan the remainder for a valid MPEG frame sync (footer-aware for v2.4) and preserve that suffix, otherwise return the input unchanged with a "refused unsafe clean" action instead of emptying the output. — L

service/scripts/av_meta.py
```diff
@@ -254,6 +268,16 @@ def _inspect_id3v2(data: bytes) -> tuple[bool, bool, list[str]]:
...
+            total = min(10 + tag_size, len(data))
+            return data[total:], [
```
bentalay, line 284: Don't drop complete-but-malformed tags via the truncation branch. — service/scripts/av_meta.py:276-278 — The branch fires on any ID3-prefixed parse failure, including full tags with bad frames (non-target frames included, even with strip_all_metadata=False); gate it on 10 + tag_size > len(data) and route non-overrun failures to the refuse path from item 1. — S

service/scripts/av_meta.py
```diff
@@ -226,6 +226,20 @@ def _parse_id3v2_frames(data: bytes) -> tuple[int, int, list[tuple[bytes, bytes]
 def _inspect_id3v2(data: bytes) -> tuple[bool, bool, list[str]]:
     parsed = _parse_id3v2_frames(data)
     if parsed is None:
+        if len(data) >= 10 and data[:3] == b"ID3":
```
bentalay, line 235: Make a failed probe inconclusive, not clean. — service/scripts/av_meta.py:229-241 — has_c2pa/has_ai still read False on an unreadable tag; return an explicit tri-state (e.g. None/"inconclusive") or a status field and propagate it through report output and exit aggregation so callers can't certify. — M

tests/test_id3_truncated.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_full_tag_still_parses_clean():
+    data = _id3_header(4) + b"abcd" + b"\x00" * 4
+    # A well-formed small tag parses without the truncation note.
+    _, _, findings = av_meta._inspect_id3v2(data)
+    assert not any("truncated" in f.lower() for f in findings)
+
+
+def test_truncated_tag_is_dropped_not_certified():
```
bentalay, line 49: Add payload-preservation tests. — tests/test_id3_truncated.py:49-56 — Add cases asserting byte-identical audio tail after cleaning: a truncated tag followed by a valid MPEG frame, a complete-but-unparseable tag with trailing payload, and an ID3v2.4 tag with footer. — M

tests/test_id3_truncated.py
```diff
@@ -0,0 +1,56 @@
...
+    data = _id3_header(160) + b"contentcredentials" + b"\x00" * 40
```
bentalay, line 50: Add a refusal/inconclusive regression test. — tests/test_id3_truncated.py — Assert the non-overrun malformed-tag path returns input unchanged with a refusal action, and that inspection reports the inconclusive status from item 3. — S

poorvith-mp:
Closing this PR.

As noted in the reviews, the stripping implementation in this PR drops the entire file down to 0 bytes whenever 10 + tag_size > len(data) (data[total:] returns empty bytes on truncated headers, causing complete data loss).

The safe, non-destructive truncated ID3v2 handling (validating MPEG frame headers, skipping false syncs, and preserving audio payloads byte-for-byte rather than wiping the file) has been implemented and tested in #201.

Thanks for the initial report and exploration here, @yzxcj797.

PR #175, opened 2026-08-19
fix: keep collected evidence when a later zip member fails to read

### LINKED ISSUE #164: [bug] one unreadable zip member discards the C2PA/AI markers already found in earlier members

yzxcj797:
## Summary

Fixes #164.

All three container inspectors (OOXML `:884`, ODT `:1351`, EPUB `:1534`) accumulate `has_c2pa`/`has_ai`/`findings` member by member, then on any member-read failure execute:

```python
except _ZIP_PARSE_ERRORS:
    return False, False, [f"not a valid {fmt.upper()} zip"], {}
```

Markers already found in earlier members are **discarded** and replaced with the hardcoded clean-looking shape. The issue's repro: a DOCX with `docProps/core.xml` carrying four markers plus one corrupt `customXml` member → `has_c2pa=False, has_ai=False, ['not a valid DOCX zip']`, and the pre-commit gate exits **0** on exactly the file that carried evidence.

The repo already holds the principle elsewhere — the zip-bomb test pins that a refused container must propagate rather than "be reported as an unparseable container" — but the partial-read case was uncovered.

## Fix — all three sites

- On a member-read failure: **keep the accumulated flags/findings**, append a note naming the partial read and the exception class (`partial read of DOCX zip (error); evidence above survives, later members were not scanned`). The C2PA/AI findings and `has_c2pa` drive `is_actionable`, so the gate still blocks.
- A **wholly-garbage container** (nothing accumulated) keeps the old `not a valid X zip` shape — the existing malformed-zip test's semantics are unchanged.
- **EPUB's except is widened** from `zipfile.BadZipFile` to the shared `_ZIP_PARSE_ERRORS` tuple, closing the asymmetry the issue documents: a zlib error crashed the scan entirely while a CRC failure discarded everything.
- EPUB's `names` is bound before the try (it was assigned inside), so a failure at archive-open can't NameError the partial branch.

## Tests

Deterministic member-read failures (monkeypatched `_read_zip_member` raising the issue's exact `zlib.error` for one member, archive structure intact — the real-world triggers: corrupt deflate, CRC mismatch, unsupported compression, encrypted member) across **all three formats**, plus the wholly-garbage and intact-container guards. The three evidence tests **fail on current `main`** (flags discarded); all 41 existing `test_container_meta.py` tests stay green.


codifly-network:
Verified this closes #164. Merged onto current `main`, my original fixture (marker in `docProps/core.xml`, one later member with a corrupted deflate stream, central directory intact):

```
before  has_c2pa=False  findings=['not a valid DOCX zip']                     check_staged exit=0
after   has_c2pa=True   findings=['docProps/core.xml: marker:c2pa, marker:C2PA, marker:contentcredentials, ai:ContentCredentials, ai:OpenAI',
                                  'partial read of DOCX zip (error); evidence above survives, later members were not scanned']
                                                                              check_staged exit=1
```

Two things I specifically tried to break and could not:

- Widening EPUB's `except zipfile.BadZipFile` to `_ZIP_PARSE_ERRORS` looked like it could flip a fail-closed path open, since `zlib.error` is in that tuple and previously propagated out to `EXIT_PARTIAL`. On an EPUB whose first scanned member raises `zlib.error` with nothing accumulated, both `main` and this branch still exit 3. No regression.
- `ZipBudgetExceeded` is correctly still outside `_ZIP_PARSE_ERRORS`, so the zip-bomb propagation pinned by `test_zip_budget_rejection_propagates_from_inspect` is unaffected.

Initialising `names: list[str] = []` before the `try` is a good catch — the EPUB except path would otherwise raise `NameError` on a failure before `namelist()` returns.

One note, not a blocker: the `if findings:` guard uses findings as a proxy for "evidence accumulated". That holds today because every site that sets `has_c2pa`/`has_ai` appends a finding in the same block, but it is an invariant nothing enforces. `if findings or has_c2pa or has_ai:` would not depend on it.

## CI

`ruff check` fails on the new test file — an unused `# noqa: E402` at line 20 (`ruff.toml` already grants `tests/**` that ignore), plus `RUF059` for unused unpacked variables at lines 56, 74, 93, 108. Baseline `main` is lint-clean, so these are new. Full suite passes.

guillaumemeyer (approved):

PR #176, opened 2026-08-19
fix: truncated ISOBMFF containers still run the C2PA byte-scan fallback

### LINKED ISSUE #167: [bug] a truncated AVIF/HEIC/MP4 skips the whole-file C2PA byte-scan its sibling inspectors run

yzxcj797:
## Summary

Fixes #167.

`inspect_isobmff` returns early when no boxes parse:

```python
boxes = _parse_isobmff_boxes(data)
if not boxes:
    return False, False, [f"not a valid {fmt.upper()} (no ISOBMFF boxes found)"]
```

That early return **skips the whole-file C2PA byte scan** at the bottom of the same function — the fallback `inspect_png`/`inspect_jpeg`/`inspect_gif`/`inspect_tiff` all reach even after a truncation. The issue's two-state repro (same AVIF, first box size intact vs overrunning, `b"c2pa"` present in **both**):

- intact → `has_c2pa=True`, `byte-scan C2PA markers: …`
- overrunning → `has_c2pa=False`, `['not a valid AVIF (no ISOBMFF boxes found)']`

Also reached for MP4/MOV/M4A through `av_meta._inspect_mp4`.

## Fix

Run the byte scan **before** reporting the parse failure:

- truncated + markers → `has_c2pa=True` with the byte-scan finding **plus** the parse-failure note (the truncation isn't hidden);
- walkable container → box-level findings unchanged (the stronger evidence — a named `c2pa`/`jumb` top-level box);
- truncated + no markers → the plain `not a valid X` shape, nothing to misreport.

## Tests

The issue's exact two-state repro plus the markerless-truncated guard:

- truncated-with-markers byte-scans to `has_c2pa=True` **and** reports the parse failure — fails on current `main` (early return reported clean);
- intact container keeps its box-level finding with no parse-failure note;
- markerless truncated keeps the old single finding.

Existing `test_avif_heic.py` + `test_av_meta.py` suites stay green (30 tests).


codifly-network:
Verified this closes #167. Merged onto current `main`, the issue's two-state fixture (same AVIF, first-box size intact vs overrunning, `b"c2pa"` present in both):

```
overrun before  has_c2pa=False  ['not a valid AVIF (no ISOBMFF boxes found)']
overrun after   has_c2pa=True   ['byte-scan C2PA markers: c2pa, C2PA, jumb, JUMB, contentcredentials',
                                 'not a valid AVIF (no ISOBMFF boxes found)']
intact          has_c2pa=True   ['byte-scan C2PA markers: ...']   (unchanged)
```

Keeping the parse-failure note alongside the byte-scan finding is the right call — the truncation stays visible instead of being replaced by the stronger result. Returning `has_ai or has_c2pa` also matches the function's normal return shape, so the early path is no longer a special case.

Worth noting for whoever picks up #170: this fixes the **inspect** side of the ISOBMFF box walk. `strip_isobmff` still rebuilds only the boxes that parsed, so a truncated AVIF is still rewritten with its `mdat` dropped. Different function, different issue, just flagging that this PR does not cover it.

## CI

`ruff check` fails on the new test file — an unused `# noqa: E402` at line 19 (`ruff.toml` already grants `tests/**` that ignore) and `RUF059` for an unused unpacked `has_ai` at line 37. Baseline `main` is lint-clean, so both are new. Full suite passes.

guillaumemeyer (approved):

PR #180, opened 2026-08-19
fix: a second --in-place run preserves the original .bak

### LINKED ISSUE #172: [bug] a second --in-place run overwrites the .bak with the cleaned output, so the original is unrecoverable
## What happened

`backup_path()` copies whatever is currently at the path, and `clean_file.py:100` calls it unconditionally on every `--in-place` run:

```python
# service/scripts/common.py:286-288
    bak = src.with_suffix(src.suffix + ".bak")
    try:
        safe_write_bytes(bak, src.read_bytes())
```

So the second run backs up the *first run's output* over the original. Its own docstring is the promise this breaks:

> Used by ``--in-place`` flows so the original is never partially lost: the
> original file stays untouched until the cleaned output is atomically
> renamed over it.

That holds for one run. After two, the original is gone.

## No corruption required

This is not tied to a damaged file or to any other bug. An ordinary Markdown file with a zero-width space, cleaned entirely correctly, twice:

```
ORIGINAL (pristine, watermarked) sha=b6cd1899907d
after run1  notes.md=3e54ef94628d  notes.md.bak=b6cd1899907d  exit=0  | original recoverable? YES
after run2  notes.md=3e54ef94628d  notes.md.bak=3e54ef94628d  exit=0  | original recoverable? NO

bak content after run2: b'# Notes\n\nHelloworld, an ordinary document.\n'
U+200B still anywhere on disk? False
```

Both runs succeeded and did exactly what they should. The pre-clean bytes simply no longer exist anywhere, and nothing said so.

## The second run is the documented path, not an edge case

`clean_staged.py` is a standard pre-commit auto-fixer: it exits non-zero whenever it modified a file so the developer reviews and re-stages, then commits again — which re-runs the hook on the same file. Rewriting files on the way into a commit is exactly where a working backup matters most, and it is the flow that reliably destroys it.

## What you expected

Either the first backup is preserved, or the user is told the backup was replaced. Right now nothing distinguishes "created a backup" from "overwrote your backup with post-clean output".

The intent is already visible in the tests. `tests/test_security_hardening.py:206` (`test_backup_path_refuses_symlinked_bak`) hardens `backup_path` so a symlinked `.bak` cannot be used to clobber an unrelated file — the function is already expected to be careful about what it destroys at that path. `tests/test_security_hardening.py:197` covers the single-run case. Nothing covers a repeat run.

## Suggested fix

Keep the first backup rather than refusing, so a legitimate re-run still cleans:

```python
    bak = src.with_suffix(src.suffix + ".bak")
    if bak.exists():
        return bak          # first backup is the pre-clean one; never replace it
```

I would avoid `raise SystemExit(2)` here even though it matches the symlink-refusal idiom at `common.py:255-256`. Exit 2 is the "unrecognized format / oversized" code that `clean_staged.py:46-49` treats as a non-fatal skip, so a refusal would turn every second hook run into a silent no-clean. (That bucket is the subject of #159 and PR #161.)

If you would rather keep every generation, unique names (`.bak`, `.bak.1`, …) also work, at the cost of sidecar litter in consumer repos — this repo's `.gitignore` covers `*.bak`, but a repo installing the hook from `.pre-commit-hooks.yaml` will not have that.

## Environment

- Reproduced in `docker run python:3.12-slim` against an unmodified checkout

## Input type

- [x] Both / batch directory
- Layer involved: A (Unicode) — but the mechanism is format-independent

## Related

Independent of #170, but it is what makes #170 permanent: when `strip_png` turns a recoverable image into an unopenable husk, the second run overwrites the one copy that could have brought it back.

yzxcj797:
## Summary

Fixes #172.

`backup_path()` copies whatever is currently at the path, and `clean_file.py --in-place` calls it **unconditionally on every run**. The second run backs up the first run's *output* over the original. The issue's repro needs no corruption — an ordinary Markdown file with a zero-width space, cleaned correctly, twice: after run 2 the `.bak` holds run 1's cleaned output and the pristine original exists nowhere.

And the second run is the **documented path**, not an edge case: `clean_staged.py` is a pre-commit auto-fixer that exits 1 whenever it modified a file (so the developer reviews and re-stages), and committing again re-runs the hook on the same file.

## Fix — preserve, don't replace

`backup_path()` returns early when a `.bak` already exists:

- the first run's pristine backup is **kept**, never overwritten;
- the return value became `(path, created)` so callers can tell "created" from "kept" — `clean_file` reports `backup notes.md.bak already exists from an earlier run; keeping the original backup`, making the two states distinguishable exactly as the issue asks;
- `clean_text.py`'s in-place path rides the same helper and inherits the fix.

## Semantic note (safe direction)

The symlinked-`.bak` security test's shape changes **for the better**: an existing `.bak` — symlink or not — is no longer written through at all, so the write-refusal path can't trigger for an existing bak. That test now pins that the symlink target stays untouched. `safe_write_bytes`' refusal for a *fresh* backup is unchanged.

## Tests

- the issue's two-run repro: first call creates (`created=True`), second preserves (`created=False`), and the `.bak` bytes are still the **pristine** original — fails on current `main`;
- fresh-backup creation unchanged;
- the updated symlink security test (target untouched).

All existing suites (`test_security_hardening`, `test_precommit_hooks`, etc.) green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Existing `.bak` backup files are now preserved instead of overwritten during in-place cleaning.
  * A warning is shown when a backup already exists.
  * Existing symlinked backups remain protected and unchanged.
  * Backup creation errors continue to be reported clearly.

* **Tests**
  * Added coverage confirming pristine backups are created once and preserved through later file changes.


guillaumemeyer:
@yzxcj797, a quick automated analysis:

Merge blocker:
- clean_image.py:132 (bak = backup_path(...); src = bak) was never updated for the new (Path, bool) return. Every --in-place image run fails with a TypeError. The issue itself says the mechanism is format-independent, and the flag is user-facing (--help: "writes .bak backup first"). No test covers the image CLI path, so CI stays green. Must-fix.

Two other findings:
- Symlink behavior change is a real defensive regression, though not a new write primitive. Base refused a pre-placed symlinked .bak (the point of test_backup_path_refuses_symlinked_bak); the PR instead reads through it and replaces the original with the symlink target's (cleaned) content, while printing a message that is factually wrong ("keeping the original backup"). Write-protection holds (victim never touched), so severity is moderate — but the test name, the message, and the semantics all need reconciling before merge.
- Dangling-symlink → exit 2 concern from the earlier report is real but adversarial-only (an attacker who can plant the symlink); the legitimate two-run flow is unaffected.

guillaumemeyer:
@coderabbitai full review

poorvith-mp (dismissed):
Ensuring `backup_path()` keeps the existing `.bak` on repeated `--in-place` runs fixes #172 and prevents the pre-commit loop from clobbering the pristine original file with cleaned intermediate output.

Security tests and backup preservation tests pass cleanly. Merged `main`. Approving.

poorvith-mp (approved):
Ensuring `backup_path()` keeps the existing `.bak` on repeated `--in-place` runs fixes #172 and prevents the pre-commit loop from clobbering the pristine original file with cleaned intermediate output.

Security tests and backup preservation tests pass cleanly. Merged `main`. Approving.

PR #184, opened 2026-08-19
fix: an empty actions list means the cleaner left the bytes alone (#173)

### LINKED ISSUE #173: [bug] every non-text clean_file run reports "changed", so the clean pre-commit hook can never pass

codifly-network:
## What

Makes an empty `actions` list mean "the cleaner left the bytes alone".

- The 18 sites that appended a `"no X removed"` filler now return `[]`.
- `clean_file.py` / `clean_image.py` print `already clean (nothing removed)` through a shared `print_actions()` helper in `common.py`, so a no-op report is still reassuring rather than blank.
- `clean_epub`'s two sentinel comparisons (`sub_actions != ["no HTML AI meta removed"]`, `!= ["no OPF metadata removed"]`) collapse to plain `if sub_actions:`.
- `strip_jpeg`'s `"preserved entropy-coded scan (SOS→EOF)"` note is now recorded only alongside a real removal.

`clean_staged._changed()` is unchanged — it becomes correct as written.

## Why

Fixes #173. `_changed()` decides "did this file change?" from `bool(result.get("actions"))`, and only the text path carries `stats` to answer it properly. Because every other kind always appended a filler, that list was never empty:

```
note.txt   -> actions=None, stats={'removed_count': 0, ...}     <- the only correct one
```

Two consequences, both fixed here. Clean, unmodified files, run twice:

```
              before          after
file        run1  run2      run1  run2
pic.png        1     1   ->    0     0
pic.jpg        1     1   ->    0     0
pic.gif        1     1   ->    0     0
pic.bmp        1     1   ->    0     0
pic.webp       1     1   ->    0     0
pic.tiff       1     1   ->    0     0
doc.md         1     1   ->    0     0
note.txt       0     0   ->    0     0
```

A repo adopting `watermarks-remover-clean` could not commit with a `.md` or an image staged: exit 1, re-stage, nothing changed, exit 1 again. And because `cleaned N file(s) in place` fired on every such commit, the one line that would announce a destructive clean (#170) was the line a developer learns to scroll past.

A file that really carries a mark still cleans and still exits 1 — verified end to end.

## Fixing the source, not the caller

`container_meta.py` already worked around its own filler:

```python
```

A caller already needed a real "nothing happened" signal and had to reconstruct it by string-matching the sentinel. `clean_staged.py` has the same need one level up without the workaround. Removing the fillers lets both sites drop the comparison.

## PDF and the image exiftool pass (third commit)

`exiftool` and `qpdf` run whether or not there is anything left to strip, and each recorded that it ran. Those records made **every PDF**, and **every image on a machine with exiftool installed**, report as modified. CI installs neither tool, so the image half of this is invisible to CI and shows up only on a developer machine.

First the clean had to be able to converge at all. qpdf regenerates the second half of the PDF `/ID` on every write, so cleaning the same PDF twice never produced the same bytes:

```
before  (--linearize)                       run1=50a207f984 run2=57a4e98a36  converges=False
after   (--linearize --deterministic-id)    run1=703dd13666 run2=703dd13666  converges=True
```

With that, `clean_pdf` and `clean_image` judge the outcome by the bytes rather than by "the tool ran". A PDF converges after one rewrite and reports nothing on every run after it:

```
clean PDF               run1: 196B -> 912B  actions=['exiftool -all= (rc=1)', 'qpdf --linearize structural rewrite (rc=3)']
                        run2: 912B -> 912B  actions=[]
PDF with AI Info dict    run1: 261B -> 876B  actions=[... the strip ...]
                        run2: 876B -> 876B  actions=[]
```

**Operation records and warnings are now separate lists.** My first attempt compared bytes and dropped the whole actions list, which silently ate `metadata bytes may remain recoverable` — a warning that the clean was *incomplete*, which is exactly what a byte comparison must never swallow. Three of your existing tests caught it. Warnings now always survive; only operation records drop, and `exiftool failed:` stays reported unconditionally.

The degraded-clean note moved from `actions` to `meta["note"]`, which `clean_file.py` now prints as a warning. Nothing printed it before — the exit code read `meta["degraded"]` while the human output relied on the action string:

```
wrote /tmp/a.out.pdf format=pdf
  - already clean (nothing removed)
  ! no PDF cleaner available (install exiftool for reliable metadata strip); copied as-is
```

Still true after this: the *first* clean of an untouched PDF rewrites it, because qpdf linearization is a real change. That is existing behaviour and the rewrite is what actually drops unreferenced objects, so I left it — adopting the hook rewrites each PDF once, then converges.

## Checklist

- [x] Docs updated (README and/or skill references) if user-facing behaviour changes — no user-facing docs describe the filler strings

## Notes for the reviewer

Layer: Files (metadata), plus the pre-commit wrapper that consumes the result.

Five existing tests asserted the filler text for an already-clean input; each now asserts `actions == []`, which pins the actual contract. New `tests/test_clean_actions_contract.py` covers nothing-removed-means-no-actions across PNG/JPEG/Markdown/HTML/SVG, that a real removal still reports, that the JPEG preserved-scan note only accompanies a removal, and the hook regression itself including idempotency. Fixtures are stdlib-only in the existing `_minimal_*` idiom — no new dependencies.

Verified in `python:3.12-slim` against the same commands CI runs: `pytest -q` green both with and without exiftool/qpdf installed (that skip is `test_pdf_structural_rewrite.py`, "needs real exiftool and qpdf", identical on `main`), `ruff check` clean, `ruff format --check` clean. I did not run the Windows or macOS legs.


codifly-network:
Closing in favour of #181, which fixes #173 from the hook side and leaves the user-facing action strings untouched. Two competing PRs for one issue is worse than either, and that is the maintainer's call to make cheaply rather than adjudicate.

Leaving a handoff, because this branch covered some things #181 does not. Measured, not assumed — `python:3.12-slim`, clean files run through the hook twice.

**Still open after #181, in CI's environment (no exiftool):**

- **JPEG** — `preserved entropy-coded scan (SOS→EOF)` (`image_meta.py:1785`) is appended to every JPEG. Exit 1, 1 with bytes identical.
- **BMP** — `no BMP trailing metadata to strip` (`image_meta.py:671`) and `BMP trailing bytes kept (keep-non-ai-metadata)` (`:675`). Same shape.

Both are one-line additions to `_NOOP_ACTION_PATTERNS` if that is the direction.

**Still open, and not really the same defect:**

- **Every PDF** gets `exiftool -all= (rc=0)` (`container_meta.py:1866`) and `qpdf --linearize structural rewrite (rc=0)` (`:1833`); **every image on a machine with exiftool** gets `exiftool -all= pass` (`image_meta.py:1986`). These describe work that was performed rather than no-op phrasings, so I do not think they belong in a no-op matcher.

- **PDF cleaning is non-deterministic**, independent of anything about `actions`. qpdf regenerates the second half of the PDF `/ID` on every write, so cleaning the same PDF twice never produces the same bytes:

  ```
  --linearize                       run1=50a207f984 run2=57a4e98a36  converges=False
  --linearize --deterministic-id    run1=703dd13666 run2=703dd13666  converges=True
  ```

  That one is worth fixing whatever happens to #173 — today every commit touching a PDF rewrites it differently.

- `clean_pdf`'s warnings and its operation records share one list, so anything that filters that list by outcome can swallow `metadata bytes may remain recoverable` — a warning that the clean was *incomplete*. I hit exactly that while building this branch and three of your existing tests caught it. Worth keeping in mind for any future filtering there.

- Nothing prints `meta["degraded"]`. `clean_file.py` reads it for the exit code, while the human output relied on the action string, so a degraded PDF clean announces itself only by accident.

Happy to reopen the `--deterministic-id` piece as its own small PR if that is useful — it is unrelated to #173 and stands alone.

PR #195, opened 2026-08-19
fix: restore the fi that bench-synthid-text swallowed

Belkins:
`d5f4f03` (#145) inserted the `bench-synthid-text` rule between the last
continued line of `smoke-markdiffusion`'s recipe and the `fi` that closed it.
Make treats the target line as part of the recipe, so both targets are broken
on current `main`:

```
$ make smoke-markdiffusion
/bin/sh: -c: line 1: syntax error: unexpected end of file
make: *** [smoke-markdiffusion] Error 2

$ make bench-synthid-text
make: Nothing to be done for `bench-synthid-text'.
```

`bench-synthid-text` is the target `docs/synthid-text-benchmark.md` tells people
to run. It has silently done nothing since 2026-08-18.

`git blame` shows it plainly — line 92's `fi` still belongs to the
`smoke-markdiffusion` block from `6144019`, with `d5f4f03`'s target wedged
above it:

```
6144019 82) smoke-markdiffusion:
6144019 86)   $(PYTHON) $(SCRIPTS)/markdiffusion_harness.py --help ...; \
d5f4f03 87) bench-synthid-text:
d5f4f03 91)   echo "run: ..."; \
6144019 92)   fi
```

## The fix

One missing `fi`.

## Keeping it from coming back

`tests/test_makefile_targets.py` checks every name in `.PHONY`:

1. a rule for it actually exists;
2. `make --dry-run` expands it to a real recipe, not `Nothing to be done`;
3. the expansion parses under `sh -n` — this is what catches a recipe that ate
   the following target;

plus a textual check that no target declaration sits inside a continued recipe
line.

I verified the guard can fail rather than assuming it: against the unfixed
Makefile it fails four ways —

```
FAILED test_phony_target_expands_to_a_recipe[bench-synthid-text]
FAILED test_phony_target_recipe_is_valid_shell[smoke-markdiffusion]
FAILED test_phony_target_recipe_is_valid_shell[bench-synthid-text]
FAILED test_no_recipe_line_swallows_the_next_target
```

— and against the fix all 94 pass. Skipped where POSIX `make` or `sh` is
unavailable, so the Windows leg of the matrix is unaffected.

Full suite green locally; `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Completed the `smoke-markdiffusion` Make target so it runs correctly and reliably.

* **Tests**
  * Added automated validation for Makefile target declarations, dry-run recipes, shell syntax, and continued recipe lines.
  * Checks run on supported POSIX environments and skip automatically when the required tools are unavailable.


guillaumemeyer:
@coderabbitai full review

poorvith-mp (approved):
Restores the missing `fi` that was swallowed between `smoke-markdiffusion` and `bench-synthid-text`.

The static target validation tests in `test_makefile_targets.py` catch target boundary slips and verify expansion under `sh -n` for all `.PHONY` recipes. Approving.

PR #198, opened 2026-08-19
Fix MP4/MOV sample offset table (stco/co64) corruption on metadata strip

### LINKED ISSUE #197: MP4/MOV metadata strip corrupts sample offset tables (stco/co64) when moov precedes mdat (faststart files)
## Bug

`av_meta.py`'s MP4/MOV cleaner (`_strip_mp4` / `_strip_moov_udta` / `strip_isobmff`
for the top-level `uuid`/`jumb` path) drops ISOBMFF boxes byte-for-byte —
`moov/udta`, top-level `uuid` (XMP/C2PA), `jumb` — without correcting the sample
offset tables (`stco` / `co64`) that live inside `moov/.../stbl`. Those tables
store **absolute file offsets** into `mdat`. Dropping any box that sits before
`mdat` shifts `mdat` by the size of the dropped box, so every `stco`/`co64`
entry now points N bytes short of the real sample data.

Every faststart-optimized file (`moov` before `mdat` — the standard way video
is packaged for web/social upload) is corrupted this way. Non-faststart files
(`moov` after `mdat`) are unaffected, because dropping `udta` doesn't move
`mdat` in that layout — which made this easy to miss in ad-hoc testing.

Worse: none of the module's checks (`has_c2pa`, `has_ai_metadata`) ever
decode a sample, so the tool reports success on a file it just corrupted.

## Reproduction

```bash
ffmpeg -f lavfi -i testsrc=duration=3:size=320x240:rate=25 -c:v libx264 -pix_fmt yuv420p \
       -metadata comment="AI generated" -movflags +faststart /tmp/fs.mp4

python3 service/scripts/clean_file.py /tmp/fs.mp4 -o /tmp/fs.clean.mp4
# reports success, metadata gone

ffmpeg -v error -i /tmp/fs.clean.mp4 -f null -
# Invalid NAL unit size (1835802669 > 3948)
# Error splitting the input into NAL units
```

Remove `-movflags +faststart` from the encode step and the cleaned output
stays valid — that's the exact discriminant: `moov` position relative to
`mdat`, not anything content-specific.

Found via three independently corrupted real-world MP4s in a downstream
integrator's pipeline, then reduced to the minimal repro above.

## Suggested fix

Don't hand-edit ISOBMFF boxes on a file whose `moov` may precede `mdat`.
Remux through `ffmpeg -map 0 -c copy [-map_metadata -1 | -metadata k=]`
instead — the muxer recomputes `stco`/`co64` from scratch (so it can't
reproduce this class of bug by construction) and, as a side effect, drops
any top-level box it doesn't understand (`uuid`/`jumb` C2PA/XMP manifests
included) since it writes a fresh container rather than copying bytes
forward. For selective (`keep_non_ai_metadata`) mode, `ffprobe -show_entries
format_tags` before the remux tells you which specific keys to null via
`-metadata k=` in the same pass — the "drop a box, then remux to fix the
offsets" approach does **not** work, because the remux demuxes the file
it's given: if that file's offsets are already desynced, the demuxer reads
packets at the wrong offsets and the remux faithfully copies forward
whatever garbage it finds there.

We've applied this fix locally (remux-based `_strip_mp4`, plus a
decode-verification guard in `clean_av()` that refuses to write output and
raises instead of returning success if cleaning regresses a file from
decodable to non-decodable — checked against the *original*'s own
decodability so non-media test fixtures don't become uncleanable). Happy to
open a PR with the diff and an added non-regression test (faststart MP4,
non-faststart MP4, WAV, MP3, keep-mode, PNG/Markdown untouched) if useful.

logicalpeyote777:
Fixes #197.

## Problem

`_strip_mp4` dropped ISOBMFF boxes (`moov/udta`, top-level `uuid`/`jumb`) by
editing bytes directly, without correcting `stco`/`co64` sample offset
tables. Any box dropped before `mdat` shifts `mdat`, desyncing every sample
offset. This corrupts every faststart-optimized file (`moov` before `mdat`
— how video is packaged for web/social) while every metadata-only check in
the module still reports success, since none of them decode a sample.

Full repro and analysis in #197.

## Fix

Replaces box-level editing for MP4/MOV/M4A/M4V with an `ffmpeg -c copy`
remux:

- `strip_all_metadata=True`: single remux with `-map_metadata -1`. ffmpeg
  rebuilds `moov`/`stco`/`co64` from scratch (can't reproduce the offset
  bug by construction) and drops any top-level box its demuxer doesn't
  understand (`uuid`/`jumb` C2PA/XMP manifests included) as a side effect
  of writing a fresh container.
- `strip_all_metadata=False` (`keep_non_ai_metadata`): `ffprobe
  -show_entries format_tags` finds which specific keys match an AI hint,
  then a single remux nulls just those via `-metadata k=`, leaving
  everything else untouched. Box-editing before a remux doesn't work for
  this case either — the remux demuxes the file it's given, so if that
  file's offsets are already desynced from a prior box edit, the demuxer
  reads garbage at those offsets and the remux just copies it forward.
  Verified this empirically before landing on the design.
- A residual-signal fallback (probe again, clear anything ffprobe's
  `format_tags` didn't originally surface, remux again) covers cases
  outside the format-tag / top-level-box mechanisms above.

Second layer, independent of the above: `clean_av()` now re-decodes every
cleaned AV file (MP4 and, as defense in depth, WAV/MP3 — their chunk/frame
stripping isn't offset-based and isn't known to have this bug, but the
check is cheap and the failure mode is severe enough to be worth covering
uniformly) before it's ever written to the destination path. It blocks on
a **regression** against the original's own decodability, not an absolute
standalone standard — so a source that was never valid media to begin with
(e.g. a hand-built test fixture) doesn't become uncleanable; only cleaning
that turns a working file into a broken one raises instead of returning
success. Lives in `clean_av()` itself, so every caller (the HTTP service,
the CLI) is covered without reimplementing it.

`server.py`'s `/capabilities` now reports `ffmpeg` (required for the MP4
family going forward).

## Tests

- `tests/test_av_meta.py`: the five existing MP4 tests used hand-built
  ISOBMFF byte structures with no real `trak`/codec data — nothing for
  ffmpeg to remux. Rebuilt on tiny real ffmpeg-encoded fixtures, same test
  intent (XMP detection+strip, udta generator-tag detection+strip,
  default-strip-all, keep-mode preserves unflagged tags, idempotent).
- `tests/test_av_corruption_regression.py` (new): the exact reported
  repro (faststart MP4) plus a non-faststart control, the selective
  keep-mode path, WAV, MP3, PNG/Markdown sanity checks (unaffected by this
  patch), and the corruption guard itself (mocked failure, proven to block
  and leave the original untouched).
- `tests/test_http_server.py`: updated `/capabilities` assertion for the
  new `ffmpeg` key.

Full suite passes locally (`pytest tests/`).

## Compatibility

`ffmpeg`/`ffprobe` become required for MP4/MOV/M4A/M4V cleaning (already an
optional dependency elsewhere in this project; not previously required for
this path). If absent, `_strip_mp4` raises a clear `ValueError` explaining
why, rather than falling back to the byte-editing path this PR removes.

## Summary by CodeRabbit

- **New Features**
  - Added `ffmpeg` to the reported available media-processing tools.
  - Improved MP4 metadata cleaning while preserving media playback and file structure.
  - Added selective metadata removal with support for retaining non-targeted metadata.

- **Bug Fixes**
  - Prevents cleaned files from being saved when processing would cause corruption.
  - Verifies cleaned audio and video remain decodable.
  - Preserves original files when validation fails.

- **Tests**
  - Expanded coverage for MP4, WAV, MP3, PNG, and Markdown cleaning workflows.


guillaumemeyer:
@coderabbitai full review

poorvith-mp (commented):
The offset table desync issue identified here is accurate: stripping a box before `mdat` shifts byte offsets and breaks `stco`/`co64` indexing.

However, the repository addressed this natively on `main` via `_isobmff_free_box()` padding in `av_meta.py` and `image_meta.py`. Instead of deleting the box and shifting subsequent bytes, dropped boxes are replaced in place with an empty `free` box of identical length. This keeps absolute byte offsets into `mdat` bit-exact without shelling out to `ffmpeg` or risking concurrency/remux artifacts.

Closing in favor of the stdlib `_isobmff_free_box` approach already in `main`. If you have specific fixtures where `free` box padding still fails playback, please open an issue with the sample.

poorvith-mp:
Closed in favor of the native stdlib _isobmff_free_box padding merged on main.

PR #200, opened 2026-08-19
fix: preserve VS16 after the emoji singletons outside the block ranges

mrsausagerolls:
## What

`_is_emoji_base` covers the emoji block ranges plus a handful of singletons, but misses the five Emoji=Yes codepoints that live outside all of them: U+203C (double exclamation), U+2049 (exclamation question), U+2139 (information source), and U+2934 / U+2935 (curved up/down arrows). A VS16 after any of these was stripped as a free-floating carrier, so a default clean visibly changed the text: U+203C U+FE0F became the bare text glyph, and ZWJ chains across them broke.

This adds the five singletons to `_is_emoji_base`, mirrors the change into the vendored skill copy (`skills/clean-user-facing-text/scripts/text_unicode.py`, byte-identity test kept green), and adds tests: VS16 preserved after each of the five, a ZWJ chain across U+2139 stays bound, and inspect reports them as not suspicious.

## Why

Layer A promises that a default clean never visibly alters correct text. These five are the only Emoji=Yes codepoints (checked against Unicode 16 emoji-data.txt) not already inside the covered ranges (0x1F000-0x1FAFF, 0x2190-0x25FF, 0x2600-0x27BF, 0x2B00-0x2BFF) or the existing singleton list, so this closes the gap exactly rather than widening any range.

Found while running a differential harness of a TypeScript port of Layer A against this engine over ~164k generated cases.

## Checklist

- [x] `python3 -m pytest -q` passes (all 81 tests, including the vendored-copy identity test)
- [x] Docs updated if user-facing behaviour changes (none beyond preserving what should never have been stripped)
- [x] No drive-by refactors

## Notes for the reviewer

Layer A only. The set was derived mechanically: enumerate every Emoji=Yes codepoint from emoji-data.txt, keep those where `_is_emoji_base` returns False; the result is exactly these five.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of additional punctuation- and symbol-based emoji.
  * Preserved emoji variation selectors so supported characters retain their intended emoji presentation.
  * Prevented valid emoji sequences, including ZWJ sequences, from being incorrectly flagged as suspicious or counted as removed.

* **Tests**
  * Added coverage for preserving punctuation and letterlike emoji sequences during text cleaning and inspection.


guillaumemeyer:
@coderabbitai full review

poorvith-mp (approved):
Correctly adds the five `Emoji=Yes` singleton codepoints (U+203C, U+2049, U+2139, U+2934, U+2935) to `_is_emoji_base`. This prevents VS16 (`\ufe0f`) following punctuation/arrow emoji from getting stripped as invisible carriers.

Mirrored cleanly into the vendored skill copy and verified against the 63-case text cleaning test suite. Merged `main`. Approving.

PR #201, opened 2026-08-20
fix: harden in-place backups, clean-file status, synthid verdict, truncated ID3v2, and zip routing

### LINKED ISSUE #163: [bug] a truncated ID3v2 tag reports zero findings, and clean_file.py then certifies the file clean

### LINKED ISSUE #165: [bug] the SynthID image score never reaches /inspect suspicious, so watermarked and scorer-down both read as clean
## What happened

`/inspect`'s `suspicious` verdict never reads the SynthID image score:

```python
# service/scripts/server.py:625-634
    detected_wm = any(
        entry.get("available") and entry.get("is_watermarked")
        for entry in report.get("text_detectors") or []
    )
    suspicious = (
        bool(report.get("suspicious_total"))
        or bool(report.get("has_c2pa") or report.get("has_ai_metadata"))
        or bool(report.get("stylometry", {}).get("score", 0.0) >= 0.65)
        or detected_wm
    )
```

`detected_wm` covers the text detectors only. `report["synthid"]` (`ImageInspectReport.to_dict`, `image_meta.py:175`) is populated and never consulted.

## Steps to reproduce

The same clean-looking PNG through `_inspect_payload`, varying only the scorer:

```
scorer says WATERMARKED    -> suspicious=False  report.synthid={'available': True, 'is_watermarked': True, 'score': 0.97}
scorer ERRORED             -> suspicious=False  report.synthid={'available': False, 'error': 'sidecar unreachable'}
no scorer configured       -> suspicious=False  report.synthid=None
```

Three distinct states, one boolean.

`inspect_image.py:68` has the same gap in its exit code:

```python
    return 0 if not (report.has_c2pa or report.has_ai_metadata) else 1
```

so the CLI prints `(watermarked: yes)` in human mode and exits 0.

## What you expected

A positive score should make the file suspicious. A scorer that failed should be distinguishable from a scorer that ran and found nothing.

`suspicious` is documented as the per-file verdict for `/inspect` and `/inspect/batch` (README:170,173; `server.py:16,261,340`), so this is the field an integrator keys on.

## Note on the rest of the detector layer

The text detectors are honest and I am not reporting them: `text_detectors.py` returns `{"available": False, "error": ...}` on timeout, exit-3, non-zero exit, bad JSON and HTTP failure, and never emits `is_watermarked: false` for a detector that did not run. The gap is specifically that the image scorer's result, positive or failed, is not wired into the verdict at all.

## Environment

- macOS (Darwin 25.5.0), arm64; commit `063119d`; scorer stubbed at `image_meta.run_synthid_score`

## Input type

- [x] Image (PNG / JPEG)

### LINKED ISSUE #172: [bug] a second --in-place run overwrites the .bak with the cleaned output, so the original is unrecoverable

### LINKED ISSUE #173: [bug] every non-text clean_file run reports "changed", so the clean pre-commit hook can never pass

### LINKED ISSUE #191: [bug] audit_website routes extensionless/generic-MIME zip containers (docx, xlsx, pptx, epub, odt) to text scanner
## What happened

In `service/scripts/audit_website.py`, `guess_kind()` identifies file types via URL extensions, Content-Type headers, and initial magic bytes. While images (PNG, JPEG, WebP, GIF, BMP, TIFF, AVIF, HEIC), PDFs, audio (WAV, MP3), SVG, and HTML have magic byte sniffers, **ZIP-based containers (DOCX, XLSX, PPTX, ODT, EPUB)** have no magic byte detection:

```python
# service/scripts/audit_website.py:188-214
    if data.startswith(b"\x89PNG"):
        return "png"
    if data.startswith(b"\xff\xd8"):
        return "jpeg"
    if data.startswith(b"%PDF"):
        return "pdf"
...
    if b"<html" in data[:2000].lower() or data[:100].lstrip().lower().startswith(b"<"):
        return "html"
    return "text"
```

When an endpoint serves a container document without an explicit file extension in the URL path (e.g., query strings like `/download?file_id=456` or `/api/export/report`) and with a generic content type like `application/octet-stream` or `application/zip`:
1. `guess_kind()` returns `"text"`.
2. `inspect_remote()` writes the bytes as `asset.txt` and calls `scan_file()`.
3. `scan_file()` executes the `text` branch, decoding binary zip data with `surrogateescape` and performing a text/Unicode watermark scan.
4. All ZIP container member inspections (`docProps/core.xml`, custom XML, embedded media C2PA manifests) are completely bypassed, reporting the asset as clean.

## Suggested Fix

Leverage `detect_container_format(Path(path), data)` or check `data[:2] == b"PK"` in `guess_kind()` to sniff ZIP container types (`docx`, `xlsx`, `pptx`, `odt`, `epub`) when extension or Content-Type is absent or generic.

---
*Note: I will open a PR with the fix and regression tests shortly.*

poorvith-mp:
### Summary

Fixes 5 defects across CLI in-place backup handling, pre-commit hook change detection, SynthID inspect verdict wiring, truncated ID3v2 metadata inspection, and ZIP container format routing.

Closes #163
Closes #165
Closes #172
Closes #173
Closes #191

### Key Changes

1. **Pre-Commit Hook & `clean_file.py` No-Op Action Handling (Fixes #173):**
   - Added `is_mutating_action` and `result_has_changes` in `common.py` to differentiate informational log entries (`"no PNG metadata chunks removed..."`) from actual mutations.
   - `clean_staged.py` and `clean_file.py` now report already-clean files as unchanged, preventing pre-commit hooks from getting stuck in re-staging loops.
2. **In-Place Backup Preservation (Fixes #172):**
   - `backup_path()` in `common.py` preserves pre-existing `.bak` copies across repeated `--in-place` runs so the original pristine copy is never destroyed.
3. **SynthID Image Score Verdict Mapping (Fixes #165):**
   - Wired `synthid` score results and findings into `inspect_image()` and `_inspect_payload()` in `server.py` so watermarked images trigger `suspicious: True` and scorer failures trigger `inconclusive` findings.
4. **Truncated ID3v2 Tag Handling (Fixes #163):**
   - `_inspect_id3v2()` in `av_meta.py` now flags truncated tags (`declared_size > remaining_bytes`) with a warning finding.
   - `_strip_id3v2()` safely strips truncated ID3v2 headers while preserving subsequent audio frame data.
5. **Extensionless ZIP Container Routing (Fixes #191):**
   - `audit_website.guess_kind()` probes ZIP magic bytes (`PK\x03\x04`) and inspects container structures (`word/`, `xl/`, `ppt/`, `content.xml`, `mimetype`, `META-INF/`) to accurately route DOCX, XLSX, PPTX, ODT, and EPUB assets without relying solely on URL extensions.

## Summary by CodeRabbit

- **New Features**
  - Improved detection of extensionless document formats, including DOCX, XLSX, PPTX, ODT, and EPUB.
  - Added detection and handling for truncated ID3v2 metadata in MP3 files.
  - Added image watermark analysis using SynthID results.

- **Bug Fixes**
  - Cleaning now clearly reports whether files were changed or already clean.
  - Existing backups are preserved during repeated in-place cleaning.
  - Improved handling of incomplete metadata and safer backup creation.


guillaumemeyer:
@coderabbitai full review

poorvith-mp:
@guillaumemeyer  All checks are passes, you can merge, or you can just comment any changes required here

poorvith-mp:
@guillaumemeyer @bentalay All CI checks across Ubuntu, macOS, and Windows matrix builds plus CodeQL and linting are passing green (713 tests passed).

This PR consolidates and hardens the fixes for:
- #163 (safe truncated ID3v2 tag inspection and audio frame sync recovery)
- #165 (SynthID score verdict mapping and probe failure signals)
- #172 (preserving `.bak` copies across repeated in-place runs)
- #173 (accurate non-mutating action detection in `clean_file.py` and `clean_staged.py`)
- #191 (extensionless ZIP container structure probing for office/epub assets)

Ready for code owner review and merge.

bentalay:
@poorvith-mp 

### 1. Every truncated ID3v2 tag becomes affirmative AI evidence — HIGH 
- **Problem:** the truncated-tag inspection branch computes the AI-marker hits but returns `has_ai=True` unconditionally, even when no marker is present.
- **Evidence:** `service/scripts/av_meta.py:241` (`return has_c2pa, True, findings`); `tests/test_truncated_id3v2.py:20-24` discards `_has_ai` without asserting it.
- **Impact:** an ordinary corrupted or cut-off MP3 is reported as containing AI metadata; `has_ai_metadata` flips in reports and `/inspect` marks clean-but-damaged files `suspicious`. This inverts the project's own principle that an inconclusive probe must never read as a verdict.


### 2. The pre-commit "changed / unchanged" decision rests on prose sniffing — HIGH 
- **Problem:** `is_mutating_action()` classifies cleaner actions by substring-matching free-form log text; that boolean drives `clean_staged`'s re-staging decision. The flaw is demonstrated in both directions:
  - **False positive (demonstrable today):** for a JPEG such as `b"\xff\xd8abc"`, `strip_jpeg()` copies the bytes verbatim and emits "copied remainder after non-marker byte" (`image_meta.py:1855-1867`) — a string absent from the denylist — so `result_has_changes()` returns true and `clean_staged` reports "changed" for an unchanged file (`clean_staged.py:39-40,85,102-104,136`), blocking the commit over a nonexistent diff: the very loop (#173) this PR claims to close.
  - **False negative (structural fragility):** denylisted words already appear in real labels emitted during mutating runs — "preserved entropy-coded scan" (`image_meta.py:1844`), "kept N bytes of truncated tail" (`image_meta.py:1999`), "… failed: …" (`image_meta.py:2068`, `container_meta.py:1859`). Correctness rests on the untested invariant that every mutating run also emits at least one non-denylisted label.
- **Evidence:** `service/scripts/common.py:280-306`; `service/scripts/clean_staged.py:40`; labels cited above.
- **Impact:** the change-detection contract lives in log-message vocabulary and can break in either direction with no test noticing. Recommended fix : a structured `mutated` flag per action, returned by the cleaners.

### 3. A single non-canonical ZIP marker now imposes a format 

- **Problem:** container detection was relaxed from marker pairs to single alternatives: `META-INF/manifest.xml` alone → ODT; any `.opf` file (or `META-INF/container.xml` alone) → EPUB.
- **Evidence:** `service/scripts/container_meta.py:188-199`; `tests/test_website_zip_container_routing.py:54-71` covers only canonical true-positive pairs.
- **Impact:** the website audit routes ordinary ZIPs to the wrong inspector; no test locks down behavior on an ambiguous archive.

### 4. The SynthID verdict predicate is defined in two places 

- **Problem:** the expression `is_watermarked or (confidence is not None and confidence >= 0.5)` is written twice.
- **Evidence:** `service/scripts/image_meta.py:1769-1771` (drives `/inspect` findings) and `service/scripts/server.py:709-714` (drives the `suspicious` flag).
- **Impact:** two sources of truth for one verdict — changing the threshold or the `None` handling on one side only makes findings and `suspicious` diverge silently; the null-confidence crash CodeRabbit reported must be patched in both places until the logic is centralized in a single helper (e.g. `synthid_is_watermarked(entry)`).

poorvith-mp:
@bentalay Good catches across the board. Addressing all 4 points:

1. **Truncated ID3v2 Verdict**: Switched `has_ai` from unconditional `True` to `bool(hits)` so corrupted/cut-off files don't trigger affirmative AI findings.
2. **Structured Change Detection**: Replaced prose sniffing in `is_mutating_action` with a structured `changed: bool` emitted by the cleaners based on direct byte/mutation comparison, eliminating string-matching false positives.
3. **Canonical ZIP Pairs**: Restored canonical indicator pairs (`META-INF/container.xml` + `.opf` for EPUB; `content.xml` + `meta.xml`/`manifest.xml` for ODT) and `mimetype` probing.
4. **Centralized SynthID Predicate**: Extracted `synthid_is_watermarked()` helper into `image_meta.py` as the single source of truth for both `image_meta` findings and `server` suspicious verdicts.

Pushing the updates and regression tests shortly.

poorvith-mp:
@bentalay All 4 review items have been resolved and verified with 720 passing tests (CI matrix green across Ubuntu, macOS, Windows, CodeQL, and lint):

1. **Truncated ID3v2 Verdict**: `_inspect_id3v2` now sets `has_ai = bool(hits)` and `has_c2pa = _classify_c2pa(hits)`. Clean-but-truncated ID3 tags report a note that metadata was unread/incomplete without flipping `has_ai_metadata` or marking files suspicious.
2. **Structured Change Detection**: Replaced prose sniffing in `is_mutating_action` with direct byte comparison (`src.read_bytes() != dest.read_bytes()`) and structured `changed: bool` flags across `clean_file.py` and `clean_staged.py`. Non-marker JPEGs and informational emissions no longer trigger false changes.
3. **ZIP Container Sniffing**: Restored canonical indicator pairs (`META-INF/container.xml` + `.opf` for EPUB; `content.xml` + `meta.xml`/`manifest.xml` for ODT) with `mimetype` precedence.
4. **Centralized SynthID Helper**: Extracted `synthid_is_watermarked()` in `image_meta.py` and unified it across `inspect_image` findings and `server.py` `/inspect` verdicts.

Ready for code owner approval and merge.

bentalay:
@poorvith-mp 

# Fix plan — PR #201

1. **Implement the `backup_path` tuple return — currently signature-only, so every `--in-place` run crashes on unpack.** — `service/scripts/common.py:322`, caller `service/scripts/clean_image.py:132` — Implement `(path, created)` with skip-if-exists logic in the body; update every `--in-place` caller in the same diff; when `created` is False, source the clean from the current file, never from the stale `.bak`. — **M**

2. **Fix `NameError` in the new ZIP routing branch: `path` is undefined in `guess_kind()`.** — `service/scripts/audit_website.py:215` — Derive a name from `url` (or pass a dummy `Path`); add a test exercising this branch end-to-end. — **S**

3. **Stop destroying files in truncated-ID3v2 stripping: no sync found returns `b""`, and the two-byte `0xFF/0xE0` scan accepts false syncs (e.g. reserved layer).** — `service/scripts/av_meta.py:276-286` — Validate full MPEG frame headers (version/layer/bitrate/samplerate, ideally confirm the next frame); when no confident audio start exists, return the original bytes with a warning action; update `tests/test_truncated_id3v2.py:41-45` (currently asserts the destructive outcome) and add a false-sync-before-valid-tail test asserting the tail byte-for-byte. — **M**

4. **Cap the `mimetype` read: naked `zf.read()` fully decompresses an attacker-controlled ZIP member (bomb risk, reachable from web downloads).** — `service/scripts/container_meta.py:190` — Read via `_read_zip_member()` under the shared cumulative budget; let `ZipBudgetExceeded` propagate and narrow `contextlib.suppress(Exception)` to decode errors only; add a compression-bomb regression test. — **S**

5. **Centralize ZIP classification instead of a private PK-magic branch calling `detect_container_format()` directly.** — `service/scripts/audit_website.py:210-217` — Register ZIP-container sniffing in `format_dispatch.classify_bytes()` and consume it from `guess_kind()`. — **M**

service/scripts/common.py
```diff
@@ -277,8 +277,50 @@ def safe_write_text(path: str | Path, text: str) -> None:
...
+        if p_in.is_file() and p_out.is_file():
+            try:
+                return p_in.read_bytes() != p_out.read_bytes()
+            except OSError:
+                pass
+    actions = result.get("actions", [])
+    return any(is_mutating_action(a) for a in actions)
+
+
+def backup_path(src: Path) -> tuple[Path, bool]:
```
bentalay, line 342: Implement the backup_path tuple return — currently signature-only, so every --in-place run crashes on unpack. — service/scripts/common.py:322, caller service/scripts/clean_image.py:132 — Implement (path, created) with skip-if-exists logic in the body; update every --in-place caller in the same diff; when created is False, source the clean from the current file, never from the stale .bak. — M

service/scripts/clean_image.py
```diff
@@ -129,7 +129,7 @@ def main() -> int:
 
     src = args.path
     if args.in_place:
-        bak = backup_path(args.path)
+        bak, _ = backup_path(args.path)
```
bentalay: Implement the backup_path tuple return — currently signature-only, so every --in-place run crashes on unpack. — service/scripts/common.py:322, caller service/scripts/clean_image.py:132 — Implement (path, created) with skip-if-exists logic in the body; update every --in-place caller in the same diff; when created is False, source the clean from the current file, never from the stale .bak. — M

service/scripts/audit_website.py
```diff
@@ -206,6 +207,14 @@ def guess_kind(url: str, data: bytes, content_type: str | None = None) -> str:
         if brand in (b"heic", b"heix", b"hevc", b"hevx", b"mif1", b"msf1"):
             return "heic"
         return "mp4"
+    if (
+        data.startswith(b"PK\x03\x04")
+        or data.startswith(b"PK\x05\x06")
+        or data.startswith(b"PK\x07\x08")
+    ):
+        ckind = detect_container_format(Path(path), data)
+        if ckind != "unknown":
```
bentalay: Fix NameError in the new ZIP routing branch: path is undefined in guess_kind(). — service/scripts/audit_website.py:215 — Derive a name from url (or pass a dummy Path); add a test exercising this branch end-to-end. — S

service/scripts/av_meta.py
```diff
@@ -254,6 +269,22 @@ def _inspect_id3v2(data: bytes) -> tuple[bool, bool, list[str]]:
 
 
 def _strip_id3v2(data: bytes, *, strip_all_metadata: bool) -> tuple[bytes, list[str]]:
+    if len(data) >= 10 and data[:3] == b"ID3":
+        major = data[3]
+        tag_size = _id3v2_size(data, 6)
+        total = 10 + tag_size
+        if total > len(data):
```
bentalay, line 302: Stop destroying files in truncated-ID3v2 stripping: no sync found returns b"", and the two-byte 0xFF/0xE0 scan accepts false syncs (e.g. reserved layer). — service/scripts/av_meta.py:276-286 — Validate full MPEG frame headers (version/layer/bitrate/samplerate, ideally confirm the next frame); when no confident audio start exists, return the original bytes with a warning action; update tests/test_truncated_id3v2.py:41-45 (currently asserts the destructive outcome) and add a false-sync-before-valid-tail test asserting the tail byte-for-byte. — M

service/scripts/container_meta.py
```diff
@@ -184,13 +185,22 @@ def detect_container_format(path: Path, data: bytes | None = None) -> str:
             try:
                 with zipfile.ZipFile(io.BytesIO(data)) as zf:
                     names = set(zf.namelist())
-                    if "word/document.xml" in names:
+                    if "mimetype" in names:
+                        with contextlib.suppress(Exception):
+                            mt = zf.read("mimetype").decode("ascii", errors="ignore").strip()
```
bentalay: Cap the mimetype read: naked zf.read() fully decompresses an attacker-controlled ZIP member (bomb risk, reachable from web downloads). — service/scripts/container_meta.py:190 — Read via _read_zip_member() under the shared cumulative budget; let ZipBudgetExceeded propagate and narrow contextlib.suppress(Exception) to decode errors only; add a compression-bomb regression test. — S

service/scripts/audit_website.py
```diff
@@ -206,6 +207,14 @@ def guess_kind(url: str, data: bytes, content_type: str | None = None) -> str:
         if brand in (b"heic", b"heix", b"hevc", b"hevx", b"mif1", b"msf1"):
             return "heic"
         return "mp4"
+    if (
```
bentalay: Centralize ZIP classification instead of a private PK-magic branch calling detect_container_format() directly. — service/scripts/audit_website.py:210-217 — Register ZIP-container sniffing in format_dispatch.classify_bytes() and consume it from guess_kind(). — M

Effort: S = small task (< 1 h) · M

poorvith-mp:
@bentalay All 5 items from the fix plan and CodeRabbit's findings are resolved and verified. CI is fully green across Ubuntu, macOS, Windows, CodeQL, and lint (726 passed).

Here is what was done:

1. **Backup tuple return and caller updates (`common.py`, `clean_image.py`, `clean_file.py`, `clean_text.py`)**: Implemented the `(bak, created)` tuple return in `backup_path`. Updated all `--in-place` callers so that when `created` is False, cleaning sources from the current file on disk instead of the stale `.bak`. Non-regular backup targets like directories and FIFOs are rejected with exit code 2 while still allowing regular file symlinks. Added regression tests in `tests/test_backup_preserved.py`.

2. **Fix `guess_kind()` URL parsing and container routing (`audit_website.py`)**: Extracted the filename safely from the URL path to eliminate the `NameError`. Routed container detection through `detect_container_format` for unified classification. Added test `test_extensionless_zip_classified_by_container_magic` in `tests/test_website_format_routing.py`.

3. **Non-destructive truncated ID3v2 stripping (`av_meta.py`)**: Added `_is_valid_mp3_frame_header` to validate version, layer, bitrate, samplerate, and emphasis fields. False sync words are skipped until a valid MPEG frame is found. If no valid frame is found, the original file is preserved instead of being wiped to empty bytes. Updated `tests/test_truncated_id3v2.py` with tests asserting byte-for-byte tail preservation and false-sync rejection.

4. **Cap `mimetype` read under cumulative zip budget (`container_meta.py`)**: Moved zip budget enforcement above `detect_container_format`. Decompression is read in chunks using `_read_zip_member`, and `ZipBudgetExceeded` propagates directly. Added regression test `test_zip_budget_rejection_propagates_from_detect_container_format_mimetype` in `tests/test_container_meta.py`.

5. **Null SynthID confidence guard (`image_meta.py`)**: Added type guards in `synthid_is_watermarked` and `inspect_image` so a null confidence from the sidecar doesn't raise a formatting error when `is_watermarked` is true. Added regression test `test_inspect_image_synthid_null_confidence_handled` in `tests/test_synthid_score.py`.

guillaumemeyer:
@poorvith-mp still some coderabbit to address + a conflict

poorvith-mp:
@guillaumemeyer Resolved the merge conflict against latest `main` in commit `d6221c5`:

1. **Clean Main Rebase & Schema Alignment**:
   - Resolved conflicts in `image_meta.py` and `server.py`.
   - Wired `synthid_is_watermarked` findings into the structured `_suspicious_report` dictionary format under the `watermark_detector` class and updated verdict assertions accordingly.
2. **CodeRabbit Checks**:
   - Verified the null SynthID confidence normalization guard in `image_meta.py`.
   - Verified non-regular backup target rejection (`.bak.is_file()`) in `common.py`.
3. **Verification**:
   - Full test suite passes cleanly locally across all formats and server endpoints.

PR is now clean, green, and conflict-free for merge.

guillaumemeyer (approved):

PR #202, opened 2026-08-20
feat: install skills into Claude Code and Cowork, ship the repo as a plugin + marketplace, and add a deterministic PostToolUse hook

codearranger:
## What

Three commits, all about distribution and triggering — no cleaning layer is touched.

**1. `install_skill.py` gains hosts.** It was Cursor-only; it now takes `--target` and `--skill`, so both shipped skills install into any supported host.

| Target | Lands in |
| --- | --- |
| `--target claude-code` | `~/.claude/skills/<skill>` (honors `CLAUDE_CONFIG_DIR`) |
| `--target claude-project --project-dir PATH` | `PATH/.claude/skills/<skill>` |
| `--target cowork` | `dist/<skill>.zip` upload bundle |
| `--target cursor` (default, unchanged) | `~/.cursor/skills/<skill>` |

Cowork, cloud, and routine sessions do not read `~/.claude/skills` on the local machine — they load the skills enabled for the claude.ai account ([docs](https://code.claude.com/docs/en/skills#skills-in-cowork-and-cloud-sessions)) — so that target builds a reproducible zip with a single top-level skill directory, for upload under **Customize → Skills → Add**. Every target validates the skill first against the packaging rules claude.ai uploads and the Skills API enforce (spec-only frontmatter, lowercase-hyphenated `name` ≤ 64 chars matching the directory, non-empty `description` ≤ 1024 chars, payload < 30 MB). Adds `--list` and `--link`; existing backup-and-replace semantics, exit codes, and Cursor defaults are untouched. `clean-user-facing-text`'s description no longer names Cursor as its only host.

**2. The repository becomes a Claude Code plugin and a single-plugin marketplace** (`.claude-plugin/`):

```
/plugin marketplace add guillaumemeyer/watermarks-remover
/plugin install watermarks-remover@watermarks-remover
```

No `skills` field, so Claude Code scans the default `skills/` directory and both skills load namespaced (`/watermarks-remover:remove-ai-marks`, `/watermarks-remover:clean-user-facing-text`). The plugin is named after the repository, not after a skill — naming it `remove-ai-marks` would produce `/remove-ai-marks:remove-ai-marks`.

**3. A `PostToolUse` hook makes cleaning deterministic.** A skill is an instruction the model may or may not follow, and the model is the thing producing the marks. A hook is executed by the harness on every matching tool call, cooperation not required. `hooks/hooks.json` registers a hook on `Write|Edit|MultiEdit|NotebookEdit` that runs `service/scripts/hook_written_file.py` against the file just written:

| Mode | Behaviour |
| --- | --- |
| `check` (default) | Reports provenance marks, leaves the file alone; exits 2 so findings reach the model. |
| `clean` | Strips them in place, then reports via `additionalContext` that the file on disk no longer matches what was written. |

Detection reuses `audit_lib.scan_file` / `is_actionable`, so the hook, the pre-commit gate, and the CI SARIF export agree on what counts as actionable; cleaning shells out to `clean_file.py`, so no cleaning logic is duplicated. Clean mode writes a sibling temp file and swaps only on a real difference, so already-clean files keep their mtime and don't retrigger file watchers. The README documents the equivalent `settings.json` wiring for non-plugin users.

## Why

The skills were installable in Grok (symlink) and Cursor only, and invocation was entirely model-elective. Claude Code and Cowork each need a different install mechanism, and the plugin/marketplace manifests then remove the install step for Claude Code users. The hook closes the remaining gap: files the agent writes get inspected whether or not the model decides to invoke a skill.

## Checklist

## Notes for the reviewer

No layer (A / B / Files) is touched — packaging, manifests, hook plumbing, and docs only. The cleaning engines and skill bodies are unchanged apart from one sentence in `clean-user-facing-text`'s frontmatter description.

**Scope limit, stated rather than papered over:** no hook can rewrite the assistant's chat message before the user reads it. Claude Code's `Stop` hook receives `last_assistant_message` read-only, and there is no pre-send filter for final responses — the same limit this project already documents for Cursor rules. The deterministic guarantee therefore covers files the agent writes, plus the pre-commit gate for anything on its way into git. Chat-only text still depends on the best-effort skill workflow.

**One wiring decision worth a look.** The hook mode is read from `CLAUDE_PLUGIN_OPTION_HOOK_MODE` / `WATERMARKS_HOOK_MODE`, *not* interpolated as `${user_config.hook_mode}` in `hooks.json`. End-to-end testing found that Claude Code 2.1.235 refuses to run a hook whose command references an option the user has never set in `/plugin manage` — a declared `default` does not satisfy it — and the hook fails with `Plugin option "hook_mode" isn't set` instead of running. Interpolating it left the hook silently dead on a fresh install, which is the worst possible failure for a guard like this. `test_hook_command_never_substitutes_an_unset_plugin_option` pins it.

Verification: full suite **576 passed, 1 skipped** (the skip is pre-existing, needs real `exiftool`/`qpdf`); `ruff check` and `ruff format --check` on `service tests` clean; `claude plugin validate . --strict` passes on both manifests. The plugin was installed end to end against a throwaway `CLAUDE_CONFIG_DIR`, and the hook was verified in a real headless session: the model wrote a file containing U+200B, the hook fired in clean mode, and the character was gone from disk without the model electing to clean anything. (An earlier revision of this description said 350 tests — that figure came from truncated output and was wrong; 576 is the measured count.)

New tests: `tests/test_install_skill.py` (21 cases), `tests/test_plugin_manifest.py` (10 cases, including the hook wiring), and `tests/test_hook_written_file.py` (20 cases — both modes, mode precedence, in-place cleaning, mtime preservation on clean files, temp-file cleanup, relative-path resolution against the payload `cwd`, notebook payloads, malformed/empty payloads, oversized files, and every payload shape that names no written file). CI needs no `claude` CLI; `make plugin-validate` is the CLI-backed check for maintainers.

Manifest URLs point at `guillaumemeyer/watermarks-remover`, matching every other link in the README, so the `/plugin marketplace add` line resolves once this lands upstream.

## Summary by CodeRabbit

* **New Features**
  * Added multi-platform skill installation for Claude Code, Cowork, and Cursor.
  * Added reproducible Cowork bundle packaging and optional symlink-based installs.
  * Added a Claude Code plugin with marketplace support and configurable watermark checking and cleaning modes.
  * Added automatic post-write watermark detection and optional cleanup.
* **Documentation**
  * Expanded installation, plugin, hook, and contribution guidance.
* **Bug Fixes**
  * Improved installation safety with validation, backups, rollback, and overwrite protection.
* **Tests**
  * Added comprehensive coverage for installers, hooks, manifests, packaging, and configuration validation.


guillaumemeyer:
@coderabbitai full review

poorvith-mp (approved):
Reviewed the full diff across \install_skill.py\, \hook_written_file.py\, the plugin manifests, and the test suite. This is ready to merge.

A few architectural details that stand out:

1. **Deterministic post-write execution over model election**: Making cleaning a \PostToolUse\ hook on \Write|Edit|MultiEdit|NotebookEdit\ solves the core failure mode of agentic provenance cleanup. Relying on an LLM to invoke a skill after editing code doesn't work consistently. Running the hook at the client tool layer guarantees every modified file gets scanned.
2. **Atomic swap and permission preservation**: The sibling temporary file write paired with \shutil.copymode\ before \os.replace\ prevents \mkstemp\ from downgrading executable permissions to \

poorvith-mp (approved):
Reviewed the full diff across `install_skill.py`, `hook_written_file.py`, the plugin manifests, and the test suite. This is ready to merge.

A few architectural details that stand out:

1. **Deterministic post-write execution over model election**: Making cleaning a `PostToolUse` hook on `Write|Edit|MultiEdit|NotebookEdit` solves the core failure mode of agentic provenance cleanup. Relying on an LLM to invoke a skill after editing code doesn't work consistently. Running the hook at the client tool layer guarantees every modified file gets scanned.
2. **Atomic swap and permission preservation**: The sibling temporary file write paired with `shutil.copymode` before `os.replace` prevents `mkstemp` from downgrading executable permissions to `0600`, while keeping mtimes untouched on clean files.
3. **Handling exit code 1 with valid JSON reports**: Distinguishing engine crashes from cleans that leave residual signals (like degraded PDFs) stops valid partial cleans from getting thrown away.
4. **Claude Code 2.1.235 option fallback**: Reading `CLAUDE_PLUGIN_OPTION_HOOK_MODE` directly from the environment instead of `${user_config.hook_mode}` interpolation prevents hooks from failing on fresh installs.
5. **Deterministic Cowork bundles**: Strict frontmatter delimiters and reproducible zip archives with normalized file modes keep the distribution targets clean.

All 627 unit tests pass, manifests validate with `claude plugin validate . --strict`, and the engine remains untouched. Approving.

PR #224, opened 2026-08-22
fix(docs): use portable base64 in the /detect example

openclawitback:
The `/detect` example in `SKILL.md` still uses `base64 -w0`, which is GNU coreutils-only. On macOS/BSD, `base64` rejects `-w`:

```
$ base64 -w0 notes.txt
base64: invalid argument notes.txt
Usage:	base64 [-Ddh] [-b num] [-i in_file] [-o out_file]
```

Since `SKILL.md` instructs the agent to build the payload exactly as written, a macOS agent host fails on the first `/detect` call.

Every other curl example in `SKILL.md` and `README.md` already uses the portable form — this one looks like it was simply missed:

- `README.md:183`, `README.md:263`
- `SKILL.md:100`, `:107`, `:140`, `:176`

This aligns the remaining one with that convention. `base64 < FILE | tr -d '\n'` produces byte-identical unwrapped output on both GNU and BSD.

### Verification

Ran the patched line verbatim against a local `make serve` on macOS 15 (Darwin 24.6.0, Python 3.14.5) — valid single-line base64, well-formed JSON, service parsed it and reported the expected findings.

Docs-only; no runtime or test behavior changes. `test_lightweight_skill_has_no_template_placeholders` is unaffected.

## Summary by CodeRabbit

* **Documentation**
  * Updated the watermark detection example to generate base64 output without relying on the GNU-specific `-w0` option.


poorvith-mp (approved):
Good catch. `-w0` is GNU-only and breaks on macOS/BSD `base64`. Standardizing on `base64 < file | tr -d '\n'` matches the rest of `SKILL.md` and `README.md`.

Merged `main` cleanly. Approving.

PR #218, opened 2026-08-21
feat: support C2PA metadata in FLAC files

yearth:
## What

- Add FLAC detection for native streams and streams prefixed with ID3v2 tags.
- Detect C2PA manifests stored in ID3v2.3/2.4 `GEOB` frames with the `application/c2pa` MIME type.
- Remove matching frames in `--keep-non-ai-metadata` mode while preserving other frame bytes, flags, and order.
- Handle ID3v2 extended headers and v2.4 footers, and remove an empty ID3 tag after cleaning.
- Document the supported FLAC carrier and explicit exclusions.

## Why

C2PA defines an ID3v2 `GEOB` frame as its FLAC manifest carrier, but FLAC files are currently not routed through the audio/video metadata cleaner. This leaves supported C2PA metadata undetected and unremoved.

## Checklist

- [x] Behaviour matches `SKILL.md` / `references/removal-matrix.md` when relevant
- [x] Unit tests updated or added under `tests/`
- [x] `python3 -m pytest -q` passes
- [x] Docs updated for user-facing behaviour
- [x] No drive-by refactors unrelated to the feature

## Notes for the reviewer

This change is limited to the Files metadata layer. It does not modify native
FLAC metadata blocks, Vorbis Comments, or waveform-domain watermarks.

The carrier was also verified around an FFmpeg-generated, decodable FLAC:
inspection reported confirmed C2PA metadata, the cleaned file passed `flac -t`,
and its native FLAC bytes were identical to the original stream.

## Summary by CodeRabbit

- **New Features**
  - Added FLAC file support.
  - Detects C2PA manifests stored in FLAC metadata.
  - Removes C2PA manifests while preserving native FLAC metadata, Vorbis Comments, and waveform watermarks.
  - Supports complete metadata removal when requested.
- **Documentation**
  - Updated supported-format documentation to include FLAC.
- **Bug Fixes**
  - Improved handling of FLAC metadata, including extended headers and optional footers.

PR #232, opened 2026-08-24
fix: reject partial ID3v2 frame parsing

yearth:
## What

- Abort ID3v2 frame parsing when a frame payload exceeds the declared tag boundary.
- Reject non-zero trailing bytes that are too short to form a frame header, while continuing to accept zero padding.
- Add FLAC keep-mode regressions proving malformed ID3v2 tags remain byte-identical.

## Why

The shared ID3v2 parser previously returned frames parsed before a malformed
suffix. For an ID3-prefixed FLAC containing a valid C2PA GEOB followed by a
truncated unrelated frame, keep-mode cleaning trusted that partial result,
removed the GEOB, and silently dropped the unparsed remainder.

Around an FFmpeg-generated, decodable FLAC, the old cleaner reduced the file
from 8,398 to 8,299 bytes by dropping the complete ID3 prefix. With this
change, the malformed input is preserved byte-for-byte and still passes
`flac -t`.

Follow-up to the late review on #218.

## Validation

- `tests/test_av_meta.py`: 36 passed
- Full pytest suite except one pre-existing exiftool-dependent PDF failure:
  passed
- The PDF failure reproduces unchanged on `origin/main`
- `git diff --check`
- Real FFmpeg FLAC byte comparison and `flac -t`

The local `pip-audit` command could not complete because temporary-environment
`ensurepip` aborted on this macOS host. This PR has no dependency changes; CI
will run the authoritative audit.

## Checklist

- [x] Behaviour matches the Files metadata preservation contract
- [x] Unit tests updated under `tests/`
- [ ] Full local pytest has one unrelated baseline PDF/exiftool failure
- [x] No documentation change needed
- [x] No drive-by refactors

## Notes for the reviewer

This change is limited to fail-closed ID3v2 frame parsing. It does not address
unsupported ID3v2.5+ tags or WAV report-field consistency.

Malformed ID3-prefixed FLAC currently falls back to the MP3 format label while
remaining byte-identical. Surfacing malformed tags as inconclusive is outside
this data-preservation fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed ID3 metadata in FLAC files.
  * Preserves original tags when truncated frames or nonzero trailing data are detected.
  * Continues removing C2PA metadata when only valid zero-padding follows the parsed frames.

### DIFF
--- a/service/scripts/av_meta.py
+++ b/service/scripts/av_meta.py
@@ -217,9 +217,11 @@ def _parse_id3v2_frames(data: bytes) -> tuple[int, int, list[tuple[bytes, bytes]
         frame_start = pos + 10
         frame_end = frame_start + frame_size
         if frame_size < 0 or frame_end > frames_end:
-            break
+            return None
         frames.append((frame_id, data[frame_start:frame_end]))
         pos = frame_end
+    if any(data[pos:frames_end]):
+        return None
     return total, major, frames
 
 
--- a/tests/test_av_meta.py
+++ b/tests/test_av_meta.py
@@ -285,6 +285,50 @@ def test_flac_ignores_truncated_c2pa_geob(tmp_path):
     assert dest.read_bytes() == data
 
 
+def test_flac_keep_mode_preserves_tag_with_truncated_frame(tmp_path):
+    truncated_title = b"TIT2" + struct.pack(">I", 20) + b"\x00\x00partial"
+    data = _flac(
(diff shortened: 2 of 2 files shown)


poorvith-mp (approved):
Good fail-closed fix. Returning None when a frame exceeds rames_end or when non-zero bytes are present in the padding area correctly prevents partial tag drops and keeps corrupted metadata untouched. Verified clean across all 701 tests.

Approved.

PR #238, opened 2026-08-25
fix: pre-commit clean hook keeps "cleaning" already-clean images and markdown forever

umran666:
If you use the `watermarks-remover-clean` pre-commit hook on any repo that stages images, Markdown, or most other non-text files, it can literally never pass. The file gets "cleaned", you re-stage it like the hook asks, commit again, and it tells you to re-stage again. Forever. Clean files that never change are reported as modified every single time.

It all comes down to how `clean_staged.py` decides whether a file actually changed. For anything that isn't plain text, it was trusting  the `actions` list in the report. But every `strip_*`/`clean_*` pretty much always appends a "nothing was removed" message to that list even when it removes nothing. So the list was never empty, which meant the hook read "changed" even when the file on disk was byte-for-byte identical.

You could watch it happen: `clean_file.py` would report `bytes_in == bytes_out`, exit 0, and the "nothing was removed" action, yet
the hook still came back with "cleaned 1 file(s), re-stage and commit again."

The image/container/audio cleaners already report `bytes_in` and `bytes_out`, so the honest signal is right there: did the cleaner actually write different bytes? `_changed()` now compares the two instead of guessing from `actions`. A file that didn't change exits 0. A file that really did change still exits 1,so the re-stage behaviour for actual edits is unchanged.

- Reproduced the bug end to end first: a clean 1x1 PNG would make the hook exit 1 and demand a re-stage. After the fix it's a clean exit 0.
- Confirmed the opposite direction too — a PNG with a real C2PA/marker chunk still gets picked up and asks for a re-stage.
- Added regression tests in `tests/test_precommit_hooks.py`, including one that runs the real `clean_file.py` subprocess on a real PNG built in-memory.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed clean-file checks incorrectly flagging byte-identical non-text files as changed.
  * Clean files now pass successfully without requiring re-staging.
  * Files with actual metadata or content changes continue to require re-staging.
  * Improved handling and reporting for skipped, oversized, malformed, crashed, interrupted, and symlink-failed cleaning operations.

* **Tests**
  * Added coverage for unchanged, modified, same-length, skipped, and oversized image and document files.
  * Expanded validation of exit codes, diagnostics, backups, manifest declarations, and cleaning failures.

### DIFF
--- a/service/scripts/clean_staged.py
+++ b/service/scripts/clean_staged.py
@@ -24,6 +24,7 @@ commit, with its traceback invisible outside pre-commit's verbose mode.
 from __future__ import annotations
 
 import argparse
+import hashlib
 import json
 import subprocess
 import sys
@@ -31,16 +32,55 @@ from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 
-from common import EXIT_PARTIAL, eprint, subprocess_creationflags
+from common import EXIT_PARTIAL, MAX_INPUT_BYTES, eprint, subprocess_creationflags
 
 CLEAN_FILE_PY = Path(__file__).resolve().parent / "clean_file.py"
 
 
+def _file_digest(path: Path) -> bytes | None:
+    """Compute the SHA-256 digest of *path* incrementally in 64 KiB chunks."""
+    h = hashlib.sha256()
+    try:
+        with path.open("rb") as f:
+            while chunk := f.read(65536):
+                h.update(chunk)
+        return h.digest()
+    except OSError:
+        return None
+
+
+def _is_modifying_action(action: str) -> bool:
+    """Return True if an action message indicates a real modification rather than filler/warning."""
+    a = action.strip().lower()
+    if a.startswith("no ") or a.startswith("warning:") or a.startswith("c2patool available"):
+        return False
+    if a.startswith("deep image pass not needed") or a.startswith("kept "):
+        return False
+    return not (" skipped" in a or " failed" in a)
+
+
 def _changed(result: dict) -> bool:
+    """Determine whether a clean_file.py JSON report indicates content was modified.
+
+    This function serves strictly as a fallback when before/after digests on
+    disk cannot be computed. Text cleaning reports explicit removal/replacement
+    statistics. For binary and container cleaners, actions that removed nothing
+    append a 'nothing was removed' filler message (issue #173), so a non-empty
+    actions list alone cannot indicate change. A change is confirmed when byte
+    counts differ, explicit change flags are set, or active modifying actions
+    (e.g. dropping chunks, blanking XMP packets) are present in the report.
+    """
+    if "changed" in result:
+        return bool(result["changed"])
     stats = result.get("stats")
     if stats is not None:
         return bool(stats.get("removed_count") or stats.get("replaced_count"))
-    return bool(result.get("actions"))
+    if "bytes_in" in result and "bytes_out" in result and result["bytes_in"] != result["bytes_out"]:
+        return True
+    actions = result.get("actions")
+    if actions:
+        return any(_is_modifying_action(a) for a in actions)
+    return False
 
 
 def _failure_detail(proc: subprocess.CompletedProcess[str], summary: str) -> str:
@@ -59,10 +99,20 @@ def _failure_detail(proc: subprocess.CompletedProcess[str], summary: str) -> str
 
 
 def _clean_one(path: Path) -> tuple[str, str]:
-    """Returns ('changed' | 'unchanged' | 'skipped' | 'failed', detail).
+    """Clean a single file in place and report its outcome.
 
+    Returns ('changed' | 'unchanged' | 'skipped' | 'failed', detail).
     *detail* explains a 'failed' status and is empty for every other status.
     """
+    try:
+        if path.stat().st_size > MAX_INPUT_BYTES:
+            eprint(f"skipping {path}: larger than {MAX_INPUT_BYTES} bytes")
+            return "skipped", ""
+    except OSError:
+        pass
+
+    before_digest = _file_digest(path)
+
     proc = subprocess.run(
         [sys.executable, str(CLEAN_FILE_PY), str(path), "--in-place", "--json"],
         capture_output=True,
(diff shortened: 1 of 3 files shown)


poorvith-mp (changes requested):
The SHA-256 before/after digest approach in `clean_staged.py` is solid—it directly checks whether the file on disk was mutated and cleanly handles same-length PDF XMP blanking without relying on action messages.

All unit tests across Ubuntu, macOS, and Windows matrix builds are passing green.

Before we can merge, please run `ruff format` to fix the single formatting check failure in `tests/test_precommit_hooks.py:18`:

```bash
ruff format tests/test_precommit_hooks.py
```

Once that's pushed and the lint job is green, this is ready to land.

umran666:
@poorvith-mp Done! Formatted `tests/test_precommit_hooks.py` with `ruff format` and pushed in commit 2a71845. All lint checks and test suites are green now. Ready for review/merge!

bentalay:
@umran666 
## Fix plan — PR #238

1. **Make the digest comparison authoritative for the changed/unchanged decision** — `service/scripts/clean_staged.py:144` — when both digests were computed, use `before_digest != after_digest` alone; fall back to `_changed(result)` only if either digest is `None` (currently `disk_changed or _changed(result)` lets a heuristic misread of an action like `preserved entropy-coded scan` force exit 1 on a byte-identical file):
   ```python
   if before_digest is not None and after_digest is not None:
       status = "changed" if before_digest != after_digest else "unchanged"
   else:
       status = "changed" if _changed(result) else "unchanged"
   ```
   **Effort: S**

2. **Add regression tests for byte-identical files whose reports contain non-filler action wording** — `tests/test_precommit_hooks.py` — mock a clean-JPEG report (e.g. `actions: ["preserved entropy-coded scan"]`, equal byte counts) and a no-op ExifTool report, assert exit 0; keep an end-to-end clean-JPEG case if a fixture is cheap to build. **Effort: S**

3. **Narrow `_is_modifying_action()` to the digest-unavailable fallback path only, and document that role** — `service/scripts/clean_staged.py` (`_is_modifying_action`, `_changed` docstring) — once fix 1 lands, the heuristic no longer needs to enumerate every filler phrasing; note in the docstring that it is best-effort for unreadable files. **Effort: S**

service/scripts/clean_staged.py
```diff
@@ -85,11 +134,18 @@ def _clean_one(path: Path) -> tuple[str, str]:
...
         return "failed", _failure_detail(proc, "clean_file.py wrote an unparsable report")
-    status = "changed" if _changed(result) else "unchanged"
+
+    after_digest = _file_digest(path)
+
+    disk_changed = (
+        before_digest is not None and after_digest is not None and before_digest != after_digest
+    )
+    status = "changed" if (disk_changed or _changed(result)) else "unchanged"
     return status, ""
```
bentalay, line 145: Make the digest comparison authoritative for the changed/unchanged decision — service/scripts/clean_staged.py:144 — when both digests were computed, use before_digest != after_digest alone; fall back to _changed(result) only if either digest is None (currently disk_changed or _changed(result) lets a heuristic misread of an action like preserved entropy-coded scan force exit 1 on a byte-identical file):

if before_digest is not None and after_digest is not None:
    status = "changed" if before_digest != after_digest else "unchanged"
else:
    status = "changed" if _changed(result) else "unchanged"
Effort: S

service/scripts/clean_staged.py
```diff
@@ -24,23 +24,62 @@
...
+    try:
+        with path.open("rb") as f:
+            while chunk := f.read(65536):
+                h.update(chunk)
+        return h.digest()
+    except OSError:
+        return None
+
+
+def _is_modifying_action(action: str) -> bool:
```
bentalay, line 52: Narrow _is_modifying_action() to the digest-unavailable fallback path only, and document that role — service/scripts/clean_staged.py (_is_modifying_action, _changed docstring) — once fix 1 lands, the heuristic no longer needs to enumerate every filler phrasing; note in the docstring that it is best-effort for unreadable files. Effort: S

tests/test_precommit_hooks.py
```diff
@@ -81,7 +124,93 @@ def test_clean_staged_already_clean_file_exits_0_unchanged(tmp_path, monkeypatch
     assert f.read_text(encoding="utf-8") == original
 
 
+def test_clean_staged_clean_image_report_exits_0(tmp_path, monkeypatch, capsys):
```
bentalay, line 127: Add regression tests for byte-identical files whose reports contain non-filler action wording — tests/test_precommit_hooks.py — mock a clean-JPEG report (e.g. actions: ["preserved entropy-coded scan"], equal byte counts) and a no-op ExifTool report, assert exit 0; keep an end-to-end clean-JPEG case if a fixture is cheap to build. Effort: S

umran666:
@bentalay Ah! I Made `before_digest != after_digest` strictly authoritative when both digests are available, keeping `_changed(result)` exclusively as a fallback if digests cannot be computed.

poorvith-mp:
@umran666 The authoritative digest check in `1e895dd` is spot on—handling `before_digest != after_digest` directly when both are present cuts right through the action string heuristics.

Two quick things to wrap this up:

1. **Rebase onto `main`:** `main` recently tagged `v0.6.0` and updated the changelog, so `README.md` is currently conflicting. A quick `git fetch upstream && git rebase upstream/main` will clear that up and let CI run on the latest commit.
2. **Finish Ben's items 2 & 3:** Add the regression test in `tests/test_precommit_hooks.py` asserting exit 0 on a byte-identical file with a non-filler report (like `actions: ["preserved entropy-coded scan"]`), and add the quick docstring note to `_changed()` stating that `_is_modifying_action` is strictly a fallback for unreadable files.

Once you push the rebase with that test, CI should be all green and we can get this landed.

umran666:
All done! 
- Rebased cleanly onto latest `main` (`v0.6.0`) and cleared the `README.md` changelog conflict.
- Added `test_clean_staged_byte_identical_with_non_filler_action_exits_0` to test exit 0 when digests match despite non-filler actions (e.g. `preserved entropy-coded scan`).
- Updated `_changed()` docstring to state it is strictly a fallback when on-disk digests cannot be computed.

poorvith-mp (approved):
Everything looks clean and well-structured now. The rebase cleared the changelog conflict, the authoritative digest check cuts right through the action heuristics, and the new regression tests verify both byte-identical files and fallback paths. All CI checks are green across macOS, Ubuntu, and Windows. Approved.

PR #242, opened 2026-08-26
fix: preserve truncated MP4 tails during metadata stripping

### LINKED ISSUE #240: clean_av on a truncated MP4 drops the media while reporting that it kept the tail
`_strip_mp4` reintroduces the failure mode that #170 and #182 fixed for PNG and ISOBMFF: on a truncated MP4 it discards everything after the last fully parsed box, while the action list says the tail was kept.

`_strip_mp4` runs `strip_isobmff` (which appends `data[scanned_end:]`, so the tail survives and the action is recorded) and then `_strip_moov_udta`, which rebuilds the file from `_parse_isobmff_boxes(data)[0]` only. The box walk stops at the first box whose declared size overruns the data, so on a partial download everything from that point on is dropped by the second pass, after the first pass had preserved it.

### Reproduction

From the repository root, against `main` (`4a0fbc3`):

```python
import struct, sys
sys.path.insert(0, "service/scripts")
import av_meta

def box(fourcc, payload):
    return struct.pack(">I", 8 + len(payload)) + fourcc + payload

moov = box(b"moov", box(b"mvhd", bytes(100)) + box(b"udta", b"    toolGenerated by OpenAI Sora"))
whole = box(b"ftyp", b"isom   isomiso2avc1mp41") + moov + box(b"mdat", bytes(4096))
truncated = whole[:-1024]          # a partial download: mdat overruns the data

cleaned, actions = av_meta._strip_mp4(truncated, strip_all_metadata=True)
print("in :", len(truncated))
print("out:", len(cleaned))
print("actions:", actions)
```

```
in : 3267
out: 187
actions: ['kept 3080 bytes of truncated tail (file truncated)', 'drop moov/udta box (generator/user-data tags)']
```

94% of the file is gone, including all of the coded media, and the action list states the opposite. `clean_av` then writes that 187-byte result over the destination and reports `bytes_out` without any indication that the media was lost.

An intact file is unaffected, so no existing test covers this: the media is only lost when the walk stops early.

### Suggested fix

`_strip_moov_udta` needs the same tail handling `strip_isobmff` already has, which is the second return value of the box walk it already calls:

```python
     actions: list[str] = []
     out = bytearray()
-    for fourcc, payload, _size, hdr in _parse_isobmff_boxes(data)[0]:
+    boxes, scanned_end = _parse_isobmff_boxes(data)
+    for fourcc, payload, _size, hdr in boxes:
         if fourcc != b"moov":
             out.extend(_build_isobmff_box(fourcc, payload, hdr))
             continue
         ...
         out.extend(_build_isobmff_box(b"moov", bytes(new_moov), hdr))
+    out.extend(data[scanned_end:])
     return bytes(out), actions
```

With that line, the same script prints `out: 3267`, and intact files come out byte-identical to before:

```
intact     strip_all=True  in= 4291 current= 4291 patched= 4291
intact     strip_all=False in= 4291 current= 4291 patched= 4291
truncated  strip_all=True  in= 3267 current=  187 patched= 3267
truncated  strip_all=False in= 3267 current=  187 patched= 3267
```

Since `_strip_mp4` already dedupes the "no MP4 metadata" actions from the first pass, the tail action from `strip_isobmff` stays the single source of the message and no duplicate is produced.

A regression test would be the truncated variant of an existing MP4 fixture asserting `len(cleaned) == len(data)` when only metadata boxes were replaced in place, mirroring the PNG and ISOBMFF cases added in #182.

Happy to send this as a PR if that is easier.

yearth:
Fixes #240

## Summary

When a truncated MP4 contains a `moov/udta` metadata box, `_strip_mp4` first preserves the unparsed tail through `strip_isobmff`, then `_strip_moov_udta` rebuilds the top-level boxes from only the fully parsed prefix. The second pass discarded the media tail while reporting that it had been kept.

This change carries the parser `scanned_end` through the `moov/udta` rewrite and appends the unparsed tail verbatim after rewriting the parsed boxes. Complete MP4 behavior is unchanged.

## Validation

- Added a regression test covering a truncated MP4 with `moov/udta` metadata and a partial `mdat` payload.
- Targeted regression test passes.
- All 37 tests in `tests/test_av_meta.py` pass.
- Full test suite passes when excluding the unrelated existing failure in `tests/test_pdf_deep_images.py::test_auto_leaves_a_marker_free_pdf_alone`.
- Ruff check, format check, and `git diff --check` pass.

Scope is limited to `service/scripts/av_meta.py` and `tests/test_av_meta.py`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cleaning truncated MP4 files now preserves trailing data exactly while removing AI metadata.
  - Truncated or incomplete media inspections are reported accurately instead of being treated as fully clean.
  - Cleanup now warns when residual metadata indicators remain and the file cannot be confirmed clean.
  - Files with inconclusive inspection retain their original contents and modification time.

- **Tests**
  - Added regression coverage for truncated MP4 files, metadata removal, byte-for-byte tail preservation, status reporting, and unchanged file handling.


bentalay:
@yearth ## Fix plan — PR #242

1. **Mark the strip result as partial when a tail is preserved unparsed, instead of silently treating the output as clean.** — `service/scripts/av_meta.py:138` — When `scanned_end < len(data)`, propagate a partial/inconclusive flag through the inspection result (`still_has_*` / `post_findings`), not just an action string; the preserved suffix may contain a truncated `moov`/`uuid` box with metadata the reinspection parser cannot see. — **M**

2. **Add a regression test with metadata inside a truncated top-level box.** — `tests/test_av_meta.py:429` — Build an MP4 whose tail is a truncated metadata-bearing box (e.g. cut mid-`moov`/`uuid`), run `clean_av`, and assert the result reports partial/inconclusive rather than clean. — **S**

tests/test_av_meta.py
```diff
@@ -426,6 +426,25 @@ def test_mp4_udta_stripped_by_default_even_without_ai_hint(tmp_path):
     assert any("udta" in a for a in result["actions"])
 
 
+def test_mp4_truncated_tail_survives_udta_stripping(tmp_path):
```
bentalay, line 429: Add a regression test with metadata inside a truncated top-level box. — tests/test_av_meta.py:429 — Build an MP4 whose tail is a truncated metadata-bearing box (e.g. cut mid-moov/uuid), run clean_av, and assert the result reports partial/inconclusive rather than clean. — S

service/scripts/av_meta.py
```diff
@@ -134,6 +135,7 @@ def _strip_moov_udta(data: bytes, *, strip_all_metadata: bool) -> tuple[bytes, l
                 continue
             new_moov.extend(_build_isobmff_box(s_fourcc, s_payload, s_hdr))
         out.extend(_build_isobmff_box(b"moov", bytes(new_moov), hdr))
+    out.extend(data[scanned_end:])
```
bentalay, line 138: Mark the strip result as partial when a tail is preserved unparsed, instead of silently treating the output as clean. — service/scripts/av_meta.py:138 — When scanned_end < len(data), propagate a partial/inconclusive flag through the inspection result (still_has_* / post_findings), not just an action string; the preserved suffix may contain a truncated moov/uuid box with metadata the reinspection parser cannot see. — M

yearth: Thanks for catching this. Addressed in e45ea53: the stripping path now propagates an inspection-incomplete state from scanned_end, clean_av reports it through still_has_ai_metadata plus an explicit "not fully inspected" post-finding, and the clean hook now surfaces unchanged-but-inconclusive results instead of silently treating them as clean. still_has_c2pa remains tied to actual detection.

tests/test_av_meta.py
yearth: Added in e45ea53. The new regression builds a metadata-bearing top-level moov/udta box truncated mid-box, verifies the bytes are preserved unchanged, and asserts that the clean result is inconclusive rather than clean. The existing truncated mdat case now checks the same reporting contract. Locally, all 38 AV tests and all 27 hook tests pass, along with the full Ruff checks.

yearth:
Hi @guillaumemeyer — I’ve rebased PR #242 onto the latest `main` (including v0.6.0). The PR is still limited to the truncated-MP4 preservation/inconclusive-reporting fix, and the focused tests, Ruff checks, and `git diff --check` pass. When you have a chance, could you please take a look? Thank you!

poorvith-mp (approved):
Clean fix for #240. The tail preservation logic carries scanned_end accurately through the moov/udta rebuild, and marking unparsed tails as incomplete inspection prevents false clean verdicts. Regression coverage across av_meta and the file hook is solid, and all CI checks are green across macOS, Ubuntu, and Windows. Approved.

PR #263, opened 2026-08-28
Pass HTTP payload bytes into inspect_* to skip a redundant read-back

### LINKED ISSUE #262: [bug] HTTP /inspect and /detect re-read the uploaded bytes after writing them to a temp file
## What happened

The HTTP handlers hold the uploaded payload in memory as `data: bytes`, write it to a temp file, and then call the inspectors — which read the same bytes back off disk. In `server.py`:

- `_inspect_payload` does `path.write_bytes(data)` and then `inspect_image(path)` / `inspect_av(path)` / `inspect_container(path)`.
- Each of those inspectors starts with `data = path.read_bytes()` (`image_meta.py`, `av_meta.py`, `container_meta.py`).

So every binary `/inspect` (and `/detect`) request reads the full payload from disk immediately after having written it from memory — a redundant read of bytes the process is already holding.

## What you expected

The inspectors should be able to reuse the in-memory bytes the handler already has, so the request does one write instead of write-then-read. The temp file is still needed (the c2patool / exiftool / SynthID subprocesses require a real path), but the read-back is avoidable.

## Steps to reproduce

1. Run the HTTP service (`python3 service/scripts/server.py`).
2. POST a binary image/av/container payload to `/inspect` (or `/detect`).
3. Trace the path: `_inspect_payload` -> `path.write_bytes(data)` -> `inspect_image(path)` -> `data = path.read_bytes()`. The same payload is written and then read within one request.

## Environment

- OS and arch: Windows
- Python version (`python3 --version`): 3.14.1
- How you run the skill (Grok skill path / symlink / scripts only): HTTP service (`server.py`)
- Optional tools present (`c2patool`, `exiftool`) and versions if relevant: not required to observe this

## Input type

- [x] Image (PNG / JPEG)
- [x] Both / batch directory

## Diagnostics

```bash
# server.py
#   _inspect_payload(data, ...):
#       path.write_bytes(data)
#       inspect_image(path) / inspect_av(path) / inspect_container(path)
#   _detect_payload(data, ...): same pattern for av / container
#
# image_meta.py / av_meta.py / container_meta.py
#   def inspect_image(path, ...):   data = path.read_bytes()   # re-reads what was just written
#   def inspect_av(path):           data = path.read_bytes()
#   def inspect_container(path):    data = path.read_bytes()

### PR DESCRIPTION
**Pull request by noQbot**
## What

Give the inspectors an optional `data=` parameter and pass the in-memory payload from the HTTP handlers, so a binary `/inspect` or `/detect` request does one write instead of write-then-read:

- `inspect_image` / `inspect_av` / `inspect_container` now accept `data=` and read the path only when it isn't supplied.
- `server.py`'s `_inspect_payload` and `_detect_payload` pass the bytes they already hold (`inspect_*(path, data=data)`), removing the redundant `path.read_bytes()` inside each inspector.

The temp file is still written — the c2patool / exiftool / SynthID subprocesses need a real path — so only the read-back is removed. The passed bytes are exactly the temp file's contents, so no report changes.

## Why

Closes #262. Every binary `/inspect` and `/detect` request wrote the uploaded payload to a temp file and then read the same bytes back off disk inside the inspector — a redundant read of data the process already had in memory. Text payloads already avoid this (they decode `data` directly).

## Checklist

- [ ] `python3 -m pytest -q` passes 
- [x] Docs updated (README and/or skill references) if user-facing behaviour
      changes 

## Notes for the reviewer

- **Layer:** Files (C2PA/metadata); HTTP binary path (`server.py`).
- **No report change:** every existing HTTP / av / container / avif / detect test passes; added a test asserting `inspect_image` honors `data=` (an empty on-disk file plus real PNG bytes proves the parse used the passed bytes, not a re-read).
- Scope is the inspect/detect path. The clean path (`clean_*(..., data=)`) is a separate change, left out here to keep this focused.
- I did not touch the 3 pre-existing `ruff UP012` warnings in `tests/test_http_server.py` (they're on `main`) — no drive-by cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when inspecting uploaded images, audio, video, and container files.
  * Media analysis now preserves and uses the original uploaded data, including for image metadata and SynthID detection.
  * Prevented inspections from depending on temporary file rereads when source data is available.
  * Maintained existing file-based inspection behavior when original data is unavailable.

* **Tests**
  * Added coverage confirming image metadata and SynthID requests use supplied image data correctly.

### CODE REVIEW

**Comment by poorvith-mp** [comment]
Clean optimization for #262. Reusing in-memory bytes across the HTTP inspection routes cuts redundant disk reads while keeping the temp file on disk for subprocess tools.

There is a small merge conflict with latest `main` because #261 just merged and added `run_synthid` and `include_exiftool` keyword-only arguments to `inspect_image`. 

To resolve the conflict in `service/scripts/image_meta.py`, combine the parameters on `inspect_image`:

```python
def inspect_image(
    path: Path,
    synthid_dir: str | None = None,
    *,
    data: bytes | None = None,
    run_synthid: bool = True,
    include_exiftool: bool = True,
) -> ImageInspectReport:
```

and update the report instantiation:

```python
    synthid=run_synthid_score(path, synthid_dir, data=data) if run_synthid else None,
```

Once you rebase or merge `main` with this resolution, we can get this approved and merged.

Issue #271, opened 2026-08-29
Capability report and repo audit disagree with what the tool actually sees


Kalivins:
Four defects found while standing the core service up and auditing two real trees with it: a 320-file manuscript folder (French and English prose plus 295 audio files) and a 439-file Next.js repo. Each one has a reproduction below. Fixes follow in three focused PRs.

Version: `main` at 474be13, core image built from `service/Dockerfile`, no optional profiles.

---

### 1. `/capabilities` reports `ffmpeg: false` on an image where ffmpeg works

`_tool_usable()` probes with `_VERSION_FLAG.get(cmd, "--version")`, and `_VERSION_FLAG` has no entry for ffmpeg. ffmpeg has no `--version`; it exits 8.

```
$ docker compose exec -T wr-core sh -lc 'command -v ffmpeg; ffmpeg -version >/dev/null 2>&1; echo "-version exit=$?"; ffmpeg --version >/dev/null 2>&1; echo "--version exit=$?"'
/usr/bin/ffmpeg
-version exit=0
--version exit=8

$ curl -s http://127.0.0.1:8765/capabilities | grep ffmpeg
    "ffmpeg": false,
```

Only the report is wrong: `clean_audio.py` and `clean_video.py` resolve the binary with `which()`. But a capability report is what a caller reads before deciding a job is impossible, so a false negative there reads as a clean verdict about a tool nobody ran.

### 2. ghostscript is advertised but never installed in the core image

`/capabilities` lists it and `_ghostscript_usable()` probes for it, yet `service/Dockerfile` does not install it.

```
$ docker compose exec -T wr-core sh -lc 'command -v gs || echo ABSENT'
ABSENT
```

Deep PDF work is therefore unavailable on an image that claims to support it.

### 3. The audit counts source files it never reads

`TEXT_EXTS` carries `.js` but none of `.ts`, `.tsx`, `.jsx`, `.mjs`, `.cjs`, and no `.gd`/`.gdshader`. `classify()` returns `unknown`, the per-file note says "unrecognized format; not scanned", and the summary still counts the file under `Files scanned`.

Witness, seven files each carrying exactly one U+200B:

```
$ audit_dir.py /data
Files scanned: 7
By kind: {'text': 3, 'unknown': 3, 'markdown': 1}
Actionable files: 4
  [probable] /data/sample.css: layer-a [zwj_family] U+200B ZERO WIDTH SPACE (Cf) x1
[...]

PR #273, opened 2026-08-29
fix(audit): scan .ts/.tsx/.gd, and align space confidence across formats

Kalivins:
Closes items 3 and 4 of #271. Two defects that meet in the same place: what a repo audit reports.

**1. The audit counted source files it never read.** `TEXT_EXTS` carried `.js` but none of `.ts`, `.tsx`, `.jsx`, `.mjs`, `.cjs`, and no `.gd`/`.gdshader`. `classify()` returned `unknown`, the per-file note said "unrecognized format; not scanned", and the summary still counted the file under `Files scanned`.

Witness of seven files, each carrying one U+200B: `.css`, `.py`, `.txt` and `.md` were caught, `.ts`, `.tsx` and `.gd` were silent. On a real Next.js repo, 184 of 439 files were counted as scanned and never read, so a zero-width carrier in a `.tsx` string passed the audit as clean.

**2. A Layer A space changed confidence with the extension.** `text_hit_confidence()` returns `informational` for `kind == "space"`; container formats go through `classify_finding_confidence()`, which had no such rule. Identical bytes came back `informational` as a `.txt` and `probable` as a `.md`, and `is_actionable()` is what `check_staged.py` reads to fail a commit.

Measured on a French manuscript: 430 correct non-breaking spaces across four chapters, every one reported as a probable AI mark because the chapters are `.md`. Those are French typography — the space before `; : ! ?` and inside `« »`. After the fix the same tree reports 4 informational findings and 0 actionable files.

Only `kind == "space"` is exempted. A real invisible carrier in markdown stays actionable, and `tests/test_space_confidence_parity.py` asserts exactly that.

Two new test files: `test_source_extensions_scanned.py` (each extension classifies as text, a carrier in each is found and actionable, an unknown binary stays unknown) and `test_space_confidence_parity.py` (container and text agree, carriers stay probable, French typography alone does not gate a commit).

`skills/clean-user-facing-text/scripts/common.py` is updated in the same commit, as `tests/test_lightweight_skill.py` requires.

One consequence worth naming: widening the extension set surfaces findings that were previously invisible, including legitimate ones. On the repo above it now reports a literal U+FEFF inside `.replace(/^﻿/, "")` in five `.mjs` files — a BOM-stripping regex written with the character rather than the escape. That is a true positive by the tool's own definition, and arguably a separate question about whether U+FEFF at offset 0 should be informational. Not touched here.

`python3 -m pytest -q` and `ruff check`/`ruff format --check` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection across JavaScript, TypeScript, JSX, and Godot source files.
  * Correctly treats non-breaking spaces as informational while preserving actionable alerts for invisible characters.
  * Maintains unknown classification for unrecognized binary files.

* **Tests**
  * Added coverage for supported source-file formats, character detection, confidence levels, and actionability.


poorvith-mp (approved):
Approved. Two audit correctness bugs fixed: .ts/.tsx/.jsx/.gd extensions were missing from TEXT_EXTS so files were counted as scanned but never read, and space confidence was inconsistent across container vs text formats causing 430 false actionable findings on legitimate French typography. Both have targeted regression tests. CI and CodeQL pass.

Issue #282, opened 2026-08-31
[bug] the audit and the pre-commit gate silently pass 42 common source, docs and i18n extensions


s1oopX:
## What happened

`TEXT_EXTS` in `service/scripts/format_dispatch.py:53` covers the JS/TS family, Python, Rust, Go and Godot. It does not cover C, C++, C#, Java, Kotlin, Ruby, PHP, Swift, Scala, Dart, shell, SQL, reStructuredText, AsciiDoc, LaTeX, or any localization resource format.

`classify()` answers `unknown` for all of those, and both consumers read `unknown` as "nothing to say":

- `check_staged.py:46` does `if item.get("kind") == "unknown": continue` — no output, no counter, no effect on the exit code. The pre-commit gate returns 0.
- `audit_lib.py:169` counts every item in `aggregate()["total"]`, which `print_human_report()` renders as `Files scanned` (`audit_lib.py:200`), while `Files skipped` only tracks read failures. `print_human_report()` emits a line only for files that have findings, so an unscanned file produces no output at all and the `"unrecognized format; not scanned"` note is never rendered.

This is the unfixed remainder of item 3 of #271. #273 added `.ts/.tsx/.jsx/.gd` and fixed the space-confidence parity; the rest of the extension table was not touched.

`.pre-commit-hooks.yaml` sets no `types`/`files` filter, so `watermarks-remover-check` does run on these files. It just has nothing to say about them. On a Java, C++, Ruby, PHP or shell repository it is a gate that always returns 0.

## What you expected

A file carrying an invisible Unicode carrier should get the same verdict whether it is named `.txt` or `.java`, and a summary that says `Files scanned: 10 / Files skipped: 0` should mean ten files were read.

## Steps to reproduce

1. Write the same U+200B carrier into ten files with different extensions:

```bash
mkdir -p /tmp/s2 && python3 - <<'PY'
import pathlib
body = "Intro​ paragraph with a hidden carrier.\n"
for n in ("a.txt","b.md","c.ts","doc.rst","main.c","App.java",
          "run.sh","q.sql","lib.rb","x.php"):
    pathlib.Path("/tmp/s2", n).write_text(body, encoding="utf-8")
PY
```

2. Audit the directory:

```
$ python3 service/scripts/audit_dir.py /tmp/s2
Files skipped: 0
Files scanned: 10
By kind: {'unknown': 7, 'text': 2, 'markdown': 1}
With C2PA: 0
With AI metadata: 0
[...]

PR #284, opened 2026-08-31
fix(audit): scan source, docs and i18n files the router walked past

s1oopX:
## What

Adds the 40 file suffixes that `classify()` answers `unknown` for today, so the Layer A scanner actually reads C, C++, C#, Java, Kotlin, Ruby, PHP, Swift, Scala, Dart, shell, SQL, single-file web components, prose markup, and localization resources.

`service/scripts/format_dispatch.py` — `TEXT_EXTS` goes from 19 to 59 entries, grouped with comments, plus a docstring on the table explaining what an omission costs. No other production file changes.

`tests/test_source_extension_coverage.py` — new regression test.

## Why

Closes item 3 of #271 for the remaining formats, and fixes the defect reported in #282.

`TEXT_EXTS` covered the JS/TS family, Python, Rust, Go and Godot. Everything else classified as `unknown`, and both consumers treat `unknown` as nothing to say:

- `check_staged.py:46` — `if item.get("kind") == "unknown": continue`. No output, no counter, no effect on the exit code, so the pre-commit gate returns 0.
- `audit_lib.py:169` — every item lands in `aggregate()["total"]`, which prints as `Files scanned`, while `Files skipped` only tracks read failures.

Ten files, each carrying exactly one U+200B:

```
Files skipped: 0
Files scanned: 10
By kind: {'unknown': 7, 'text': 2, 'markdown': 1}
Actionable files: 3
```

Only `.txt`, `.md` and `.ts` were read. `.pre-commit-hooks.yaml` sets no `types`/`files` filter, so on a Java, C++, Ruby, PHP or shell repository `watermarks-remover-check` was a gate that always returned 0.

#273 fixed this for `.ts/.tsx/.jsx/.gd`; this is the same defect for the rest.

## Checklist

- [x] `python3 -m pytest -q` passes — 908 passed, 139 skipped, rebased onto `main` with #201
- [x] Docs updated (README and/or skill references) if user-facing behaviour changes — no document enumerates the extension table (grepped `*.md` for existing entries such as `.gdshader`/`.mjs`; no hits), so there was nothing to update

## Notes for the reviewer

**Layer involved:** A (Unicode). No fixtures added — the tests write into `tmp_path`.

**Two suffixes deliberately left out** of the 42 that currently classify as `unknown`:

- `.env` — reading env files by default deserves its own decision, not a drive-by.
- `.m` — Objective-C, MATLAB or Mercury depending on the repo.

**Collision check.** None of the 40 appears in `CONTAINER_EXTS`, `IMAGE_EXTS` or `AV_EXTS`. `test_extension_tables_stay_disjoint` pins that invariant for all four tables so a future addition cannot quietly steal a suffix from another pipeline.

**Negative control.** With `format_dispatch.py` reverted and the new test file kept, the suite fails as expected — `test_extension_classifies_as_text` and `test_same_carrier_same_verdict_as_txt` across every parametrized name, plus `test_precommit_gate_catches_source_files`. The test is a real guard, not a tautology.

**UTF-16.** Raised in review, and worth stating plainly: `audit_lib.scan_file()` reads text as UTF-8 with `surrogateescape`, so a carrier inside a UTF-16 file is missed — on `.txt`, `.csv` and `.json` today, exactly as much as on the suffixes added here. Widening the table inherits that gap rather than introducing it, and still improves on the status quo where these files were not read at all. Two tests cover the behaviour: `test_utf16_localization_file_yields_no_findings` (widening cannot invent findings) and `test_utf16_handled_the_same_as_the_pre_existing_text_extensions` (new and pre-existing suffixes agree, and will keep agreeing once a BOM-aware decoder lands). Happy to send that decoder as a separate PR — it changes the read path for every text file plus the `/inspect` upload path, so it seemed to want its own review. See the thread on `format_dispatch.py` for the measurements.

**`test_clean_source_file_still_passes`** covers the reverse concern: ordinary source with no carrier stays clean.

**The accounting half of #282 is not in this PR**, on purpose — `Files scanned` counting unread files, and `check_staged.py` staying silent on `unknown`. Those touch `aggregate()` and `print_human_report()`, which are shared with `audit_website.py`, and `tests/test_audit.py:143` asserts on the summary shape. Happy to send it as a separate PR if you want it.

## Summary by CodeRabbit

- **New Features**
  - Expanded text-file recognition to cover additional programming, scripting, markup, documentation, localization, configuration, and tabular-data formats.
  - Unlisted file extensions remain unrouted.

- **Bug Fixes**
  - Improved consistency when validating source and localization files, including UTF-16 content.
  - Preserved correct handling of image, container, antivirus, and text file classifications.


service/scripts/format_dispatch.py
```diff
@@ -50,6 +50,11 @@
     ".markdown",
     ".mdx",
 }
+#: Plain-text formats the Layer A scanner reads. An extension that is absent
+#: here, and whose bytes sniff as neither image nor container nor av,
+#: classifies as "unknown" — audit_lib skips it and check_staged passes it
+#: without a word, so a carrier in an unlisted extension is silently missed.
+#: Keep this table broad: every entry only has to decode as text.
```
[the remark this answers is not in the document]
s1oopX: Real gap, but it belongs to the text pipeline rather than to this table, so I have pinned it rather than fixed it here.

`audit_lib.scan_file()` reads every `"text"` file as UTF-8 with `surrogateescape`, so a carrier inside a UTF-16 file is missed on the suffixes that were already in `TEXT_EXTS`, not just the new ones. Same U+200B payload, current `main`:

| file | encoding | suspicious | actionable |
| --- | --- | --- | --- |
| `a.txt` | utf-8 | 1 | True |
| `a.txt` | **utf-16** | **0** | **False** |
| `b.csv` | utf-16 | 0 | False |
| `c.json` | utf-16 | 0 | False |
| `Localizable.strings` | utf-8 | 1 | True |
| `Localizable.strings` | utf-16 | 0 | False |
| `s.ps1` | utf-8 | 1 | True |
| `s.ps1` | utf-16 | 0 | False |

`.txt`, `.csv` and `.json` have been in the table since well before this PR and behave exactly like the two suffixes flagged here. So widening the table neither introduces the gap nor deepens it — it inherits it, and it still strictly improves on the status quo, where these files were not read at all.

The suggested fix — a shared BOM-aware decoder used by both `audit_lib.scan_file()` and `server._inspect_payload()` — changes how every text file in the project is read. That is a real behaviour change for images-adjacent callers and the HTTP surface, and CONTRIBUTING asks for no refactors unrelated to the fix, so I would rather not fold it into an extension-table PR.

What I added instead (d99dc12) is a parity test:

```python
def test_utf16_handled_the_same_as_the_pre_existing_text_extensions(tmp_path):
    verdicts = {}
    for name in ("legacy.txt", "rows.csv", "Localizable.strings", "deploy.ps1"):
        path = tmp_path / name
        path.write_text(CARRIER, encoding="utf-16")
        verdicts[name] = is_actionable(scan_file(path))
    assert len(set(verdicts.values())) == 1, verdicts
```

I deliberately did not write a test asserting that the UTF-16 carrier *is* detected, nor one asserting it is *not*: the first fails until the decoder lands, and the second would enshrine the bug. The parity assertion is true today and stays true the moment a BOM-aware decoder is added, since every suffix flips together.

@guillaumemeyer happy to send the decoder as a separate PR if you want it — it is a contained change, but it touches the read path for every text file and the `/inspect` upload path, so it seemed worth its own review rather than riding along here.

service/scripts/format_dispatch.py
```diff
@@ -50,6 +50,11 @@
     ".markdown",
     ".mdx",
 }
+#: Plain-text formats the Layer A scanner reads. An extension that is absent
+#: here, and whose bytes sniff as neither image nor container nor av,
+#: classifies as "unknown" — audit_lib skips it and check_staged passes it
```
guillaumemeyer: **[suggestion]** The new `#:` docstring comment (right-side lines 53-57) is verbose and embeds design/architecture rationale that is drift-prone: it names two consumers and their exact behavior ("audit_lib skips it and check_staged passes it without a word, so a carrier in an unlisted extension is silently missed"). This restates the routing decision and documents implementation history rather than giving a concise WHY. (Note: the existing `#:` style for `CLASSIFY_HEADER_BYTES` is why-driven; the consumer-behavior narration here is the excess.)

**Suggestion:** Trim to a single WHY line such as `#: Plain-text formats read by the Layer A scanner; an absent suffix routes as "unknown" and is left unread. Keep this table broad.` and leave the classification consequence in the PR body.

tests/test_source_extension_coverage.py
```diff
@@ -0,0 +1,167 @@
...
+
+``TEXT_EXTS`` in ``format_dispatch.py`` grew one contributor at a time: it
+carried ``.gd``/``.gdshader`` for Godot and the whole JS/TS family, but none of
+C, C++, Java, Kotlin, C#, Ruby, PHP, Swift, shell, SQL, reStructuredText, or any
+localization resource format. ``classify()`` answered ``unknown`` for those, and
+both consumers read ``unknown`` as nothing to say: ``audit_lib`` records
+"unrecognized format; not scanned", and ``check_staged.py`` skips the file
+without printing a line, so the pre-commit gate returned 0.
+
+Witness before the fix — ten files, each carrying exactly one U+200B::
```
guillaumemeyer: **[suggestion]** The module docstring (lines 1-21) is a long change narrative ending in an embedded past-run transcript ("Witness before the fix — ten files, each carrying exactly one U+200B::" with the by-kind output). This documents architecture history and narrates the before/after rather than explaining why the regression test exists, so it will drift as the tables change.

**Suggestion:** Condense to one short paragraph stating the invariant ("a carrier's verdict must not depend on file extension; extensions outside the JS/TS/Python family used to classify `unknown` and be silently skipped"). Keep the `Witness before the fix` transcript in the PR description.

service/scripts/format_dispatch.py
```diff
@@ -70,6 +75,53 @@
...
+    # Prose and documentation markup
+    ".rst",
+    ".adoc",
+    ".asciidoc",
+    ".org",
+    ".tex",
+    # Localization resources: where user-facing strings actually live
+    ".po",
+    ".pot",
+    ".strings",
```
guillaumemeyer: **[suggestion]** `.strings` is added under the rationale "Localization resources: where user-facing strings actually live", but the canonical macOS `.strings` file is UTF-16, and `audit_lib.scan_file()` decodes text with `utf-8`/`surrogateescape` (confirmed). The author's own `test_utf16_localization_file_yields_no_findings` / `test_utf16_handled_the_same_as_the_pre_existing_text_extensions` pin that a carrier inside a UTF-16 file is missed. So for the most common `.strings` payload, the new entry routes the file to a scanner that still cannot surface a carrier — the localization motivation is largely unrealized for that format (UTF-8 `.po`/`.pot`/`.properties`/`.arb` do benefit). This is the pre-existing pipeline limitation, not a new defect, and the tests acknowledge it.

**Suggestion:** Either note explicitly that `.strings` coverage is UTF-8-only until a BOM-aware text decode lands, or document that the entry mainly guards UTF-8 `.strings`. Do not imply the localization extensions close the gap for UTF-16 files.

service/scripts/format_dispatch.py
```diff
@@ -70,6 +75,53 @@
     ".yml",
     ".toml",
     ".csv",
+    # Compiled, JVM and systems languages
+    ".c",
```
guillaumemeyer: **[suggestion]** Cross-module side effect of widening `TEXT_EXTS`: `classify()` is shared, so `clean_file.py` (and via it the `watermarks-remover-clean` hook in `clean_staged.py`) now routes these 40 suffixes as `text` and will rewrite them with `clean_text()` (normalize_spaces defaults on) instead of the previous `unknown` → "refusing to classify" (exit 2 → `skipped`). For UTF-16 `.strings`, `guard_binary` still refuses (it sees NUL bytes), so no corruption; for UTF-8 source/`.po` files a space homoglyph or soft hyphen will now be replaced. This is an intentional consequence of the fixing goal, but it changes what the clean hook modifies in source repos and is easy to miss from the diff.

**Suggestion:** Confirm the widening of the clean path is intended and call it out in the PR body; if the clean hook should stay conservative about source code, consider routing the clean decision separately from the audit classification.

service/scripts/format_dispatch.py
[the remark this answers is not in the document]
guillaumemeyer: Yes please create a dedicated PR for the decoder

guillaumemeyer:
Thanks for the PR — the extension-widening is correct, and pinning the UTF-16 gap with the parity test rather than folding a decoder in here is the right call.

To close out the review comments, I've prepared a small follow-up that is **comment/docstring only — no behavior change**. Could you rebase it into the branch?

**Commit:** `e958536` — `docs(audit): tighten TEXT_EXTS and coverage-test comments per review`

It addresses the review suggestions:
- Collapses the verbose `#:` docstring on `TEXT_EXTS` down to a concise WHY line.
- Replaces the "Witness before the fix" transcript in the coverage-test docstring with the invariant it actually pins (the transcript stays in the PR description).
- Scopes the localization comment to the UTF-8 `.po`/`.pot`/`.arb` case and notes that UTF-16 `.strings` still needs the BOM-aware decoder.
- Calls out that `TEXT_EXTS` also drives the clean/rewrite hook path, so it isn't audit-only.

Verified: `test_source_extension_coverage.py` plus `test_format_dispatch.py`, `test_precommit_hooks.py`, `test_clean_file_clean_exit.py`, `test_source_extensions_scanned.py` — **102 passed, 0 failures**.

The commit is on this repo's `review/pr-284-fixes` branch. To cherry-pick it:

```bash
git fetch https://github.com/guillaumemeyer/watermarks-remover review/pr-284-fixes
git cherry-pick e958536
```

Or apply the patch below with `git am`:

```
From e958536ef2819d7e179b740f7c385871ec6989e1 Mon Sep 17 00:00:00 2001
From: guillaumemeyer <guillaumemeyer@users.noreply.github.com>
Date: Mon, 31 Aug 2026 09:07:31 -0700
Subject: [PATCH] docs(audit): tighten TEXT_EXTS and coverage-test comments per
 review

- Collapse the verbose #: docstring to a concise WHY line and note that
  TEXT_EXTS also drives the clean/rewrite hook path.
- Scope the localization comment to UTF-8 (.po/.pot/.arb); UTF-16 .strings
  still needs the BOM-aware decoder.
- Replace the coverage-test module docstring's narrative and "Witness before
  the fix" transcript with the invariant it pins.
---
 service/scripts/format_dispatch.py      | 12 ++++++------
 tests/test_source_extension_coverage.py | 25 +++++--------------------
 2 files changed, 11 insertions(+), 26 deletions(-)

diff --git a/service/scripts/format_dispatch.py b/service/scripts/format_dispatch.py
index f6a3571..8740101 100644
--- a/service/scripts/format_dispatch.py
+++ b/service/scripts/format_dispatch.py
@@ -50,11 +50,9 @@ CONTAINER_EXTS = {
     ".markdown",
     ".mdx",
 }
-#: Plain-text formats the Layer A scanner reads. An extension that is absent
-#: here, and whose bytes sniff as neither image nor container nor av,
-#: classifies as "unknown" — audit_lib skips it and check_staged passes it
-#: without a word, so a carrier in an unlisted extension is silently missed.
-#: Keep this table broad: every entry only has to decode as text.
+#: Plain-text formats the Layer A scanner reads and the clean hooks rewrite. An
+#: absent suffix is left unread and unrouted, so keep this table broad: every
+#: entry only has to decode as text.
 TEXT_EXTS = {
     ".txt",
     ".text",
@@ -110,7 +108,9 @@ TEXT_EXTS = {
     ".asciidoc",
     ".org",
     ".tex",
-    # Localization resources: where user-facing strings actually live
+    # Localization resources: where user-facing strings actually live. Text is
+    # read as UTF-8, so this covers UTF-8 .po/.pot/.arb; UTF-16 .strings still
+    # needs the BOM-aware decoder.
     ".po",
     ".pot",
     ".strings",
diff --git a/tests/test_source_extension_coverage.py b/tests/test_source_extension_coverage.py
index b73aa94..848ee4c 100644
--- a/tests/test_source_extension_coverage.py
+++ b/tests/test_source_extension_coverage.py
@@ -1,23 +1,8 @@
-"""An invisible carrier must not change verdict with the file extension.
-
-``TEXT_EXTS`` in ``format_dispatch.py`` grew one contributor at a time: it
-carried ``.gd``/``.gdshader`` for Godot and the whole JS/TS family, but none of
-C, C++, Java, Kotlin, C#, Ruby, PHP, Swift, shell, SQL, reStructuredText, or any
-localization resource format. ``classify()`` answered ``unknown`` for those, and
-both consumers read ``unknown`` as nothing to say: ``audit_lib`` records
-"unrecognized format; not scanned", and ``check_staged.py`` skips the file
-without printing a line, so the pre-commit gate returned 0.
-
-Witness before the fix — ten files, each carrying exactly one U+200B::
-
-    Files skipped: 0
-    Files scanned: 10
-    By kind: {'unknown': 7, 'text': 2, 'markdown': 1}
-    Actionable files: 3
-
-Only ``.txt``, ``.md`` and ``.ts`` were read; the summary still counted all ten
-as scanned. On a Java, C++, Ruby, PHP or shell repository the
-``watermarks-remover-check`` hook was a gate that always returned 0.
+"""A carrier's verdict must not depend on the file extension.
+
+Extensions outside the JS/TS/Python family used to classify as ``unknown``,
+which both consumers read as nothing to say; these tests pin that a source or
+localization file is read the same as ``.txt``.
 """
 
 from __future__ import annotations
-- 
2.43.0
```

s1oopX:
Rebased e958536 into the branch as d286e44 and pushed.

poorvith-mp (approved):
LGTM. Widening `TEXT_EXTS` to cover compiled, scripting, shell, and localization formats ensures the Layer A scanner and clean hooks don't walk past source and resource files. Tests pin extension classification, pre-commit failure behavior, and UTF-16 parity cleanly.

poorvith-mp (approved):
Reviewed and approved.

**Summary:** Expands TEXT_EXTS from 19 to 59 entries so the Layer A scanner actually reads C, C++, C#, Java, Kotlin, Ruby, PHP, Swift, Scala, Dart, shell, SQL, single-file web components, prose markup, and localization resources. Without this, watermarks-remover-check was a gate that always returned 0 on non-JS/TS/Python/Rust/Go repos — completely ineffective.

**Quality notes:**
- Two extensions deliberately left out with clear rationale (.env — security decision; .m — ambiguous ObjC/MATLAB/Mercury)
- 	est_extension_tables_stay_disjoint prevents future suffix collisions across all four tables
- Negative control proves tests aren't tautologies
- UTF-16 gap inherited (not introduced), with decoder planned as separate PR
- No drive-by refactors — one production file, one table change

**Tests:** Parametrized coverage across all 40 new extensions, collision invariant, clean-file pass-through, UTF-16 handling parity. Suite: 908 passed, 139 skipped.

Excellent first contribution.

PR #280, opened 2026-08-30
feat(bench): SynthID recipe search + robust measurement

guillaumemeyer:
## What

Turns the SynthID-text benchmark into a **recipe search** that answers the question: *given a rendered document, what is the best combination of Layer B strategies, each at a specific intensity, that clears the mark at the highest robust rate while preserving meaning and reading human?* It also hardens the measurements the search depends on.

## Why

The old benchmark only produced a per-variant table. A single strategy pass can clear by a hair (margin ≈ 0.02) and still be reported as "cleared", and a `backtranslate` that the backend never actually applied showed up as a misleading **0% clear with lex div ≈ 0.02**. Neither tells you the best *recipe*, and neither is trustworthy for ranking.

## Changes

**`rewrite_text.py`**
- **No-op guard:** a rewrite that changed fewer than `--noop-lex-floor` (default 0.05) bigrams is flagged `noop` in `--json-stats` with a warning. A near-verbatim output can no longer be read as "0% clear".
- **Strength × intensity:** `--rewrite-level` now *modulates* the named strength prompt instead of replacing it, so `(strength, intensity)` steps are expressible for composition.

**`bench_synthid_text.py`**
- Surface `noop` and `robust_cleared` per row; aggregate `noop_n` / `robust_clear_rate`; add `robust %`, `human ↓`, `AUROC ↓` and `noop` columns to report.md and results.csv.
- **AUROC:** population-level post-removal AUROC (rewritten-watermarked vs rewritten-plain scores) plus a baseline sanity AUROC, via a rank-based Mann-Whitney implementation.
- **Human-likeness:** pluggable `--human-backend` (`stylometry` always on; optional offline `lastde`/`binoculars` detector; degrade gracefully). Human-likeness is a gauge, not proof of authorship.
- **Recipe mode (`--mode recipe`):** parse/validate `strength@intensity` recipes, compose a recipe end-to-end, Phase 1 per-strength intensity sweep + Phase 2 per-weight-vector beam search, **dominance-based (weight-independent) Pareto frontier**, and a "Best recipes" report section plus `recipes.csv`. Recommends one recipe on that frontier; `--layer-a-after` defaults **off** (rewrite backend assumed watermark-safe).

**Docs/Makefile:** document the new measurements and recipe mode; add `bench-full` and `bench-semantic` targets.

## Measurement honesty

- Same-config MarkLLM detection only; **not** Google's production SynthID-Text keying (retired from the API Aug 2026). The report keeps saying so.
[...]

Issue #186, opened 2026-08-19
bench_synthid_text.py: sanity-gate exclusion rate for morphologically rich (non-English) generation


lizzy-606:
Great methodology in #145 — the sanity gate (excluding samples where the watermark isn't detected pre-removal) is the right way to keep the clear-rate honest.

One open question about scope: the bundled corpus and MarkLLM generation path run on opt-1.3b, which is English-only. For morphologically rich / inflectional languages (Polish, Finnish, Turkish, etc.), tokenization fragments word forms far more heavily than in English, which could weaken the SynthID signal before any removal attempt — i.e., at the generation/detection stage itself, not at the rewrite stage.

If that's right, the sanity-gate exclusion rate (samples dropped because the watermark was never reliably detected to begin with) should be measurably higher for inflectional-language generation than for English, independent of anything Layer B does.

Has this been tested, even informally? If not, it might be worth a follow-up variant: run the same pipeline with a multilingual or Polish-capable base model and compare pre-removal exclusion rates across languages. That would separate two very different claims — "the watermark is robust and Layer B breaks it" vs. "the watermark was never reliably embedded for this language class in the first place" — which the current English-only benchmark can't distinguish.

(Non-native speaker — Claude helped with translation/phrasing on this one.)

poorvith-mp:
@lizzy-606 This is a spot-on observation.

In morphologically rich languages, subword fragmentation increases the token count per semantic unit and alters the n-gram context distributions that SynthID's PRF hash functions depend on. If token transitions are split across subwords, the watermark signal can become diluted or require longer generation lengths to reach confident detection thresholds.

This is exactly why the sanity gate exists in `bench_synthid_text.py`. It separates pre-removal embedding strength from post-removal clearance rate, so we never falsely claim a watermark was removed if it was never reliably embedded in the first place.

The benchmark harness is already built to evaluate this directly without any codebase changes:
1. You can supply a multilingual generator model via `--generator-model` (for example, `Qwen/Qwen2.5-7B` or `Llama-3.1-8B-Instruct`).
2. You can pass a non-English prompt dataset via `--corpus-dir`.
3. The benchmark outputs (`results.json` and `report.md`) explicitly log `sanity_gate_filtered` and `mean_score_before`, isolating pre-removal embedding failures from Layer B clearance rates.

[...]

PR #294, opened 2026-08-31
fix(bench): validate weight grid and widen the recipe search spectrum

guillaumemeyer:
## Intro — the 20-min window limitation

This benchmark runs a **live MarkLLM generation/detection pass** plus a rewrite backend (a model loaded and invoked per call), so a full search over the corpus can take a very long time. The environment enforces a **wall-clock tolerance of roughly ~20 minutes** per run: anything longer may be killed before it writes results. That is exactly what happened to the prior full run — it was **killed by the environment mid-search after ~1.5 h** and produced no results. The current run was therefore relaunched at a smaller, bounded size (4 docs × 2 seeds, 3 intensities, 4 weight vectors, beam 3, max-passes 2) to land inside the ~20 min window.

This window is the central constraint on this change. The old default weight grid was **invalid**, and the old order-only search was **too narrow** to answer the motivating question; the fix and the redesign must widen the spectrum **without blowing past the runtime budget**. That is why the search now exposes a `--phase2-levels-per-strength` cap.

## Objective — “can this mark be removed at 300 tokens?”

The run answers a concrete question: **can this mark be removed at 300 tokens?** (short-document robustness). To answer it the search must cover the aggressive end of the recipe space — higher intensities, more aggressive multi-step recipes, and a realistic `--target-margin` for a short doc. If the tightened run still can't clear the mark, the honest finding is that at 300 tokens the **mark resists these attacks**, which the report now states explicitly instead of leaving an empty frontier to interpret.

## What changed

Follow-up to #280 (recipe search). **`bench_synthid_text.py`**
- **Weight-grid fix:** `--weight-grid` default `0.33/0.33/0.33` → `0.34/0.33/0.33`. The old vector sums to 0.99 and was rejected by `parse_weight_grid` *only after* the expensive generation pass had already run, wasting the run. Now the grid/spec/recommend-weight are parsed and validated **up front** (fail fast) before generation.
- **Recipe-search redesign:** Phase 2 now ranks each weight vector's top `--phase2-levels-per-strength` intensities per strength and runs an **intensity × order** beam search (each strength used at most once), instead of freezing every strength to one intensity picked under a single hardcoded weight. This reaches high-intensity and aggressive multi-step recipes.
- **Recommended recipe:** picked from the weight-independent **Pareto frontier** under `--recommend-weight` (default `0.5/0.3/0.2`), replacing the old hardcoded-weight global winner that could be off-frontier.
- **Honest verdict:** `recipe_search` returns a `verdict` (`removable` / `partial` / `resists` / `undetermined`); the report renders a **Verdict** section so “the mark resists the searched attacks at this token length” is explicit.

New flags: `--phase2-levels-per-strength` (default 3) and `--recommend-weight` (default `0.5/0.3/0.2`), both wired into the config/command string and validated in the fail-fast path.

**Docs:** `docs/synthid-text-benchmark.md` updated to describe the intensity × order search, the new flags, the verdict, and the budget lever.

## Tests

Added: `parse_weight_vec` validation, `_weighted_score`, `_best_on_frontier` (recommend shifts with weight), `_recipe_verdict` (removable/partial/resists/undetermined), the `resists` report rendering, and a monkeypatched `recipe_search` smoke that confirms intensity variance and no strength re-use. The pre-existing `0.33/0.33/0.33` regression test is retained. Full suite + `ruff` pass.

## Runtime note

Widening the spectrum costs more evals per run. To stay inside the ~20-min window, shrink docs×seeds, `--intensity-grid`, `--weight-grid`, `--beam`, `--max-passes`, or `--phase2-levels-per-strength`.

## Summary by CodeRabbit

* **New Features**
  * Expanded recipe search to explore strength orderings and per-strength intensities.
  * Added configurable search breadth and recommendation weights.
  * Recommendations now balance multiple criteria and provide fallback results when needed.
  * Reports show removable, partial, resistant, or undetermined verdicts.
  * Added clear reporting when no recipe meets watermark criteria.
  * Added validation for search settings and saved configuration details for reproducibility.

* **Documentation**
  * Updated recipe-search guidance, recommendation behavior, verdicts, and runtime-budget controls.

### DIFF
--- a/docs/synthid-text-benchmark.md
+++ b/docs/synthid-text-benchmark.md
@@ -305,11 +305,16 @@ applied sequentially - each step's output feeds the next.
   produces the per-strength intensity curves (robust %, sem div, human_like vs
   intensity).
 - Phase 2 runs a beam search (`--beam`, `--max-passes`) once per weight vector in
-  `--weight-grid`, appending the best single steps. The scalarized objective
-  (`0.5 removal / 0.3 meaning / 0.2 human`) only drives the search.
+  `--weight-grid`, combining an order of strengths with the top
+  `--phase2-levels-per-strength` intensities for that weight, so both step order
+  and intensity are explored.
 - The report's **Pareto frontier** is computed by dominance (no weights) over the
   union of all candidates, so it is weight-independent. The "recommended" recipe
-  is one highlighted point on it.
+  is the frontier point best matching `--recommend-weight` (a
+  w_removal/w_semantic/w_human triple summing to 1.0, default `0.5/0.3/0.2`).
+  If no recipe in the searched space clears the mark, the report says the mark
+  **resists** the searched attacks at that token length rather than leaving an
+  empty frontier to interpret.
 - `--recipes "chunk@0.6,paraphrase@0.3"` composes and scores one explicit recipe
   instead of searching.
 - `--layer-a-after` re-runs the Unicode scrub on the final output; default **off**
@@ -324,4 +329,6 @@ applied sequentially - each step's output feeds the next.
 
 A recipe search is expensive (each candidate = a full rewrite chain per sample).
 Run Phase 1 coarsely first with fewer docs/seeds, then confirm the winning
-recipes on a larger run with adequate statistical power.
+recipes on a larger run with adequate statistical power. Lower
+`--phase2-levels-per-strength` (or cut docs/seeds, `--beam`, `--max-passes`) to
+keep a run inside a tight wall-clock budget.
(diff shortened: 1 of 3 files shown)


lizzy-606:
@guillaumemeyer  Heads up — this recipe-search weight-grid validation inherits a pre-existing
issue relevant here: `_lexical_divergence` in `rewrite_text.py` tokenizes with
`re.findall(r"[A-Za-z0-9]+", ...)`, ASCII-only. For any language with
non-ASCII letters (Polish, German, etc.), words get shattered at each
diacritic before the bigram-Jaccard score is computed, so lexical_divergence
— one of the three axes this PR's weight grid optimizes against — is
systematically noisy for those languages. Flagged with repro in #186;
happy to open a PR for the one-line fix if useful.

PR #301, opened 2026-08-31
ci: fail build when optional backend requirements cannot resolve

guillaumemeyer:
## Problem

The CtrlRegen, MarkLLM, and MarkDiffusion backends are installed only by their own `setup_*.sh` / `Dockerfile.*`, which CI never builds. The CI matrix installs only `requirements-dev.txt` (+ `requirements-synthid-scorer.txt` for `pip-audit`).

As a result, a bad pin in one of those optional requirement files — e.g. the transformers bump in dependabot #300, which changed `transformers` to 4.57.6 and made `pip install -r requirements-markllm.txt` (or `requirements-ctrlregen.txt`) fail with `ResolutionImpossible` against the pinned `huggingface_hub` — passed CI with all checks green. That breakage is only detectable by actually resolving each requirement file.

## Change

Add a `deps` job that runs `pip install --dry-run` against each optional backend's requirement files. `--dry-run` resolves the full dependency graph without downloading or installing anything, so it costs seconds and fails the build on a resolution conflict.

- `requirements-ctrlregen.txt`
- `requirements-markllm.txt` + `requirements-semantic.txt` (semantic is documented to be installed into the MarkLLM venv)
- `requirements-markdiffusion.txt`

`pip-audit` is intentionally NOT used here: those research pins (diffusers 0.27.2, transformers 4.37.2, ...) carry known, accepted CVEs, so an audit would fail the build with expected findings. The goal of this job is purely to catch dependency-resolution regressions.

## Verification

- Validated the workflow YAML parses and the `deps` job is registered.
- Confirmed `pip install --dry-run` exits 0 on the current baseline for all three commands (Python 3.12).
- Confirmed the same command exits 1 (fails) against the #300 change (`transformers==4.57.6`), proving this job catches the bug that previously slipped through.

## Note

This PR does not change runtime behavior — it only makes CI stricter, so #300 (and similar future Dependabot bumps) will show a red dependency check instead of passing silently.

## Summary by CodeRabbit

* **Chores**
  * Added CI validation for optional research dependencies.
  * Dependency resolution failures are now reported as CI failures.
  * Validation checks requirements without downloading packages.


guillaumemeyer:
Addressed both CodeRabbit findings in commit `d312fe7` (rebased onto updated `main`, which now includes the merged Dependabot PRs #297/#298/#299):

- **persist-credentials** — set `persist-credentials: false` on the `deps` job checkout. The job performs no authenticated git operations, so persisting the token is unnecessary exposure; the workflow only grants `contents: read`.
- **`--ignore-installed`** — added to all three `pip install --dry-run` commands so each requirements set resolves independently of packages already on the runner.

Validated: the three commands exit 0 on the current baseline and exit 1 on a #300-style broken pin (`transformers==4.57.6`). The `deps` and `lint` checks pass on the updated head.

PR #308, opened 2026-09-01
fix(image): cap decompressed PNG zTXt/iTXt size

guillaumemeyer:
## Summary

Closes the gap reported in GHSA-cfhj-799j-rm4c (thanks @kurisu725).

PNG `zTXt` and compressed `iTXt` were inflated with `zlib.decompress()` and no output cap. Zip members (`MAX_ZIP_DECOMPRESSED_BYTES`) and sitemap gzip already have budgets; this path did not. `MAX_BODY_BYTES` / `MAX_INPUT_BYTES` only bound the *compressed* payload, so a small PNG could expand to hundreds of MB–GB during inspect (and several more copies in the marker scan).

## Change

- Decode compressed PNG text through `zlib.decompressobj().decompress(..., max_length=)` with `MAX_PNG_TEXT_DECOMPRESSED_BYTES` (1 MiB). PNG text is metadata; that is well above any legitimate Software/parameters chunk.
- Over-budget chunks raise `PngTextBudgetExceeded` instead of allocating the expansion.
- `inspect_png` records `not fully inspected (decompressed text exceeds cap)` (informational, same bucket as truncated chunks) and keeps walking later chunks.
- `strip_png(..., strip_all_text=False)` drops the chunk rather than expanding it to decide whether it is AI metadata. Default `strip_all_text=True` already dropped text chunks without decompressing.

Reachability is unchanged: `inspect_file` / `audit_dir` / `audit_website` / `POST /inspect` all go through `inspect_png`. Default `/clean` was already safe; `keep_non_ai_metadata` is the clean path that now drops bombs.

## Tests

- Over-budget `zTXt` / `iTXt` are refused and do not surface markers that exist only after inflate (tiny monkeypatched cap, no large fixtures).
- Under-budget compressed text is still scanned.
- A bomb chunk does not skip a later `tEXt` generator tag.
- Selective clean drops the over-budget chunk.

`python -m pytest -q` and `ruff check` / `ruff format --check` on `service` + `tests` are green locally.

## Advisory

Private report: https://github.com/guillaumemeyer/watermarks-remover/security/advisories/GHSA-cfhj-799j-rm4c

This is a hardening fix (CWE-409 / CWE-400), same class as the zip-bomb work. Default bind is still loopback; I would not treat this as a CVE-worthy High unless the HTTP service is documented as internet-facing. After merge, accept/credit the advisory and skip or delay a CVE unless you want one for completeness.


kurisu725:
patch looks good, thanks

PR #328, opened 2026-09-05
fix(av): route .ogg/.opus/.aac to the audio chain instead of "unknown"

### LINKED ISSUE #327: [bug] .ogg/.opus/.aac route to unknown — the audio chain lists them but classify_bytes can't reach them
`clean_audio.py` lists `.ogg`, `.opus` and `.aac` in `AUDIO_EXTS` and maps all three to codecs in `_CODEC_BY_EXT`. The comment above the set says it covers them on purpose. But `AV_EXTS` in `av_meta.py:44` is `{".mp4", ".mov", ".m4a", ".m4v", ".wav", ".mp3", ".flac"}`, and that's the set `classify_bytes()` actually routes on. All three land in `unknown` and get refused before the audio chain ever runs.

The magic-byte fallback doesn't cover it either. `detect_av_format()` sniffs ISOBMFF, RIFF/WAVE, fLaC and ID3/MPEG frame sync, but never `OggS`, so an extension-less Ogg is invisible twice over.

On `d9e9590`:

```
a.ogg    classify=unknown   is_audio_name=True  codec=libopus
a.opus   classify=unknown   is_audio_name=True  codec=libopus
a.aac    classify=unknown   is_audio_name=True  codec=aac
a.m4a    classify=av        is_audio_name=True  codec=aac
no-extension OggS bytes -> unknown
```

Through the CLI it surfaces as:

```
refusing to classify t.ogg: unrecognized format
These bytes match no supported text, image or container format.
Pass --force-text to handle them as text anyway, or --as to force a format.
```

That message suggests `--force-text` for an audio file and never mentions audio or video, so there's nothing pointing at the real cause.

Practical effect: the destructive transform chain from #266 can't be applied to Opus or AAC, which covers most of what comes out of a browser recorder or a Discord export. `/clean` returns a 400 for the same inputs.

Happy to send the fix. Add the three extensions to `AV_EXTS`, add an `OggS` branch to `detect_av_format()`, and widen the refusal message to name audio and video. I'd keep it to exactly that — Ogg page-header parsing for metadata inspection is a separate question and I'd rather not fold it in.

poorvith-mp:
﻿Closes #327.

## What

- Adds `.ogg`, `.opus` and `.aac` to `AV_EXTS`, so `classify_bytes()` routes them to the audio chain that already knows how to handle them.
- Adds an `OggS` capture-pattern branch to `detect_av_format()` so an extension-less Ogg is detected from its bytes.
- Adds an ADTS branch (`0xFF 0xF1` / `0xFF 0xF9`) **above** the existing MPEG frame-sync check. Raw AAC satisfies the sync mask, so without that ordering every ADTS file would report as MP3.
- Widens `AUDIO_FORMATS` to agree with `AUDIO_EXTS`.
- The classify refusal message now names audio and video.

## Why

`clean_audio.py` already lists all three extensions and maps them to codecs, with a comment saying so deliberately. `AV_EXTS` never got them, so they died in `classify_bytes()` before the chain ran. The destructive transform from #266 was unreachable for Opus and AAC, which covers most browser-recorder and Discord audio.

The old error told the user to try `--force-text` on an audio file and didn't mention audio at all, so nothing pointed at the real cause.

## Scope

Routing only. No Ogg page-header parsing, no Vorbis comment inspection, no metadata work — the audio chain handles these formats fine once it's reached.

## Tests

`tests/test_audio_routing.py`. Includes the ADTS-vs-MPEG ordering guard and an invariant that `AUDIO_EXTS` and `AV_EXTS` can't drift apart again. Full suite green, ruff clean.

## Summary by CodeRabbit

- **New Features**
  - Added recognition for OGG, OPUS, and AAC audio formats through file extensions and signatures.
  - AAC and OGG receive distinct format classifications, while MPEG frame synchronization continues to identify MP3 files.
  - Expanded audio format handling to include OGG and AAC.
  - OGG and AAC files can now pass through cleaning unchanged.

- **Bug Fixes**
  - Updated unsupported-format guidance to identify audio and video files.

- **Tests**
  - Added coverage for audio routing, detection, passthrough behavior, and messaging.

### DIFF
--- a/service/scripts/av_meta.py
+++ b/service/scripts/av_meta.py
@@ -41,13 +41,13 @@ from image_meta import (
     strip_isobmff,
 )
 
-AV_EXTS = {".mp4", ".mov", ".m4a", ".m4v", ".wav", ".mp3", ".flac"}
+AV_EXTS = {".mp4", ".mov", ".m4a", ".m4v", ".wav", ".mp3", ".flac", ".ogg", ".opus", ".aac"}
 
 
 @dataclass
 class AVInspectReport:
     path: str
-    format: str  # mp4 | wav | mp3 | flac | unknown
+    format: str  # mp4 | wav | mp3 | flac | ogg | aac | unknown
     has_c2pa: bool
     has_ai_metadata: bool
     findings: list[str] = field(default_factory=list)
@@ -66,7 +66,7 @@ class AVInspectReport:
 
 
 def detect_av_format(data: bytes) -> str:
-    """Sniff MP4/MOV/M4A/M4V (ISOBMFF), WAV, MP3, or FLAC from magic bytes."""
+    """Sniff MP4/MOV/M4A/M4V (ISOBMFF), WAV, MP3, FLAC, OGG, or AAC from magic bytes."""
     if len(data) >= 12 and data[4:8] == b"ftyp":
         return "mp4"
     if len(data) >= 12 and data[:4] == b"RIFF" and data[8:12] == b"WAVE":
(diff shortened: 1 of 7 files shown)


guillaumemeyer:
@poorvith-mp please review coderabbit comments

PR #339, opened 2026-09-12
fix(audit): count only the files a scanner actually read

Yurakonoplya:
## What

`aggregate()` splits `total` into `read` and `unrecognized`, `print_human_report()`
reports `Files scanned` as the read count and adds a `Files unrecognized` line,
`audit_dir.py`'s JSON `files_scanned` follows the same definition, and
`check_staged.py` names the files it did not read instead of dropping them
with a bare `continue`.

Exit codes are unchanged on every path.

## Why

This is the accounting half of #282; #284 closed the extension-table half.

A file whose suffix no table claims is never opened, but `aggregate()` counted
it in `total`, which `print_human_report()` renders as `Files scanned`, while
`Files skipped` only tracks read failures. Nine files, one U+200B carrier each,
four of them with suffixes the tables know:

```
Files skipped: 0
Files scanned: 9
By kind: {'text': 3, 'markdown': 1, 'unknown': 5}
```

Five files were never read, and the summary says otherwise. `check_staged.py`
had the same hole from the gate side: `if item.get("kind") == "unknown": continue`,
so the hook returned 0 without a word about files it never opened.

Widening the extension table moves individual suffixes across that line, but
whatever stays outside it keeps getting reported as scanned. This changes what
the numbers mean instead, so the guarantee holds for suffixes nobody has
thought of yet.

The repository audits itself as an example. Before:

```
Files scanned: 279
```

After:

```
Files scanned: 267
Files unrecognized: 12
```

The twelve are `.gitignore`, `.dockerignore`, `.env.example`, `LICENSE`,
`Makefile`, five `Dockerfile*` and `integrations/cursor/clean-user-facing-text.mdc`.

## Checklist

- [x] `python3 -m pytest -q` passes: 1340 passed, 35 skipped
- [x] Docs updated (README and/or skill references) if user-facing behaviour changes: no document quotes the summary lines (grepped `*.md` for `Files scanned` / `files_scanned` / `Files skipped`: no hits), so there was nothing to update

## Notes for the reviewer

**Layer involved:** A (Unicode), audit and pre-commit surfaces only. No fixtures
added; the tests write into `tmp_path`.

**Exit codes.** Deliberately untouched. An unscanned file is a gap in coverage,
not a finding, so no existing gate turns red because the hook learned to speak.
`audit_dir.py` still returns `EXIT_PARTIAL` only for read failures.

**Counted, not listed.** `Files unrecognized` prints a number, matching the
existing `Files skipped` line. The paths and their `"unrecognized format; not
scanned"` notes stay in `--json`, so a repository full of binaries does not turn
the human report into a file listing. Happy to add a `--list-unrecognized` flag
if you would rather have them inline.

**`summary["total"]` is unchanged** (`read + unrecognized`), so anything reading
it keeps working; `tests/test_audit.py::test_aggregate_summary` still asserts
`total == 2` and passes untouched.

**`files_scanned` in `audit_dir.py --json`** now counts read files rather than
`len(files)`, because the CI surface carried the same claim as the human one.
`tests/test_audit_sarif_and_concurrency.py` asserts `files_scanned == 12` on a
tree of ten `.txt`, one `.png` and one `.md`, all of which are read, so the
number is the same and the test passes untouched.

**Negative control.** With the three production files reverted and the new test
kept, five of the six tests fail. The sixth
(`test_precommit_gate_stays_quiet_when_every_file_was_read`) passes both before
and after on purpose: it pins that the new stderr note stays silent on a tree
that was fully read.

**What this does not do.** It does not decide whether an unrecognized file
should be read (sniffing, option (b) in #282), and it does not widen any table.
It only stops the tools from claiming they read what they did not.

### DIFF
--- a/service/scripts/audit_dir.py
+++ b/service/scripts/audit_dir.py
@@ -129,7 +129,9 @@ def main() -> int:
     summary = aggregate(files)
     report = {
         "root": str(root),
-        "files_scanned": len(files),
+        # Not len(files): "files" also holds the items no pipeline claimed, and
+        # those were never opened. Their count is summary["unrecognized"].
+        "files_scanned": summary["read"],
         "files_skipped": skipped,
         "summary": summary,
         "files": files,
--- a/service/scripts/audit_lib.py
+++ b/service/scripts/audit_lib.py
@@ -164,9 +164,17 @@ def is_actionable(item: dict[str, Any]) -> bool:
 
 
 def aggregate(files: list[dict[str, Any]]) -> dict[str, Any]:
-    """Build the summary block shared by directory and website audits."""
+    """Build the summary block shared by directory and website audits.
+
+    "total" is every item the audit considered; "read" is the subset a scanner
+    actually opened, and "unrecognized" the rest. They are separate numbers
+    because a file no pipeline claimed carries no verdict: reporting it as
+    scanned turns "no findings" into "clean" for content nothing looked at.
+    """
     summary = {
         "total": len(files),
+        "read": 0,
+        "unrecognized": 0,
         "by_kind": {},
         "with_c2pa": 0,
         "with_ai_metadata": 0,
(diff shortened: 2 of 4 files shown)


poorvith-mp (approved):
LGTM. Aligns cleanly with the core engine vision and resolves the audit accounting half of #282 without breaking gate exit codes.

PR #345, opened 2026-09-14
Frontmatter value scan deletes Claude Code agent tool grants

jjlopezruiz:
## The problem

`inspect_markdown()` / `clean_markdown()` use **the same pattern for frontmatter keys and values**, and that pattern (`AI_META_NAME_RE`) matches bare vendor names — including `claude`.

A Claude Code subagent definition carries its tool grant in frontmatter:

```yaml
---
name: inbox-router
description: "Triages the inbox and routes each mail to its project."
tools: Read, Write, Edit, mcp__claude_ai_acme__mail__get-message
model: opus
---
```

The substring `claude` inside the MCP tool name makes `tools` a **value hit**. `clean_markdown()` drops the **whole key** on a value hit, so cleaning that file **deletes the agent's tool grant** — the file still loads, the agent just silently loses its tools. `model:` has the same problem via `AI_FRONTMATTER_KEYS`.

This is reachable through the plugin's own `PostToolUse` hook: it reports the file as carrying provenance marks and suggests `clean_file.py --in-place`, which is what causes the damage. It affects any `.claude/agents/*.md` or `SKILL.md` in any repository.

Found when the hook flagged a real agent definition; the suggested remedy would have removed the MCP grant that file documents as load-bearing.

## The fix

Keys and values get different patterns, which is the distinction the code was missing:

| Position | Pattern | Rationale |
|---|---|---|
| **Key** | `AI_FRONTMATTER_KEYS` + `AI_META_NAME_RE` | A key is a *claim about the document*. `claude:`, `generator:`, `made_with:` are provenance whatever the value. Bare vendor names belong here. |
| **Value** | new `AI_META_VALUE_RE` | A value is *content*. Naming a vendor says nothing about how the file was made. Only provenance-shaped phrasing counts: `ai-generated`, a watermark scheme, or an explicit `generated with <vendor>`. |

Two follow-on details:

1. **Closing the key-side gap.** Narrowing the value pattern removed the only thing catching keys like `generated-with:` and `made_with:`, whose *values* are bare vendor names. `AI_META_NAME_RE` now also matches `(generated|created|made|written|produced|authored)[-_ ]?(with|by|using)`, so those fire on the key — where they always should have.

2. **`model`/`tools` exemption, by shape not by path.** These are legitimate provenance keys in an ordinary document (`model: gpt-4` in a generated report *is* provenance) and configuration only in an agent definition. Since `inspect_container()` never receives a path, the file is recognised by its shape: `name` **and** `description` **and** (`tools` **or** `allowed-tools`) — all three required. Ordinary prose frontmatter does not carry the set.

   **Values are still scanned inside an agent definition**, so `description: Generated with Claude Code` is still caught and still stripped. That is tested explicitly, since the exemption is exactly where someone could try to smuggle a mark past the scanner.

`clean_markdown()` mirrors `inspect_markdown()` in both respects, so the two still predict each other.

## Tests

New `tests/test_frontmatter_false_positives.py` — 12 cases covering both directions in one file deliberately: the false positives that must stop firing, and every real watermark that must keep firing.

- agent `tools:` naming a vendor is not provenance
- agent frontmatter survives `clean_markdown()` byte-identical
- prose merely mentioning a vendor is not provenance
- `ai_generated:`, `generator:`, `generated-with:`, `made_with:`, `written-by:`, `c2pa:`, and `model: gpt-4` (non-agent frontmatter) all still flagged
- a watermark *inside* an agent definition is still caught and stripped
- two of the three shape keys is not an agent definition and gets no exemption

Full suite passes (`python3 -m pytest -q`), `ruff check` clean.

## Also

`docs/frontmatter-scanning.md` records the key/value asymmetry and the bug that motivates it, so the next person to widen one of these patterns sees why they are not the same pattern.

No behaviour change outside markdown frontmatter scanning.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented legitimate agent configuration metadata from being incorrectly flagged or removed during Markdown inspection and cleanup.
  - Reduced false positives when vendor names appear in tool grants, model values, or ordinary prose.
  - Continued detecting and removing genuine AI-generated metadata and embedded watermark indicators.
  - Expanded recognition of common provenance labels and descriptions.

- **Documentation**
  - Added guidance explaining frontmatter scanning behavior, agent configuration exceptions, and provenance detection rules.


poorvith-mp (dismissed):
LGTM. Strongly aligns with project vision and agent plugin reliability.

- Prevents a critical false positive where bare vendor matching in frontmatter values stripped Claude Code agent tool grants and model configurations.
- The structural shape validation (
ame + description + 	ools/llowed-tools) correctly scopes the exemption without opening holes for real watermarks.
- Comprehensive coverage in 	ests/test_frontmatter_false_positives.py ensures both directions (no false positives, all true watermarks retained) hold.

Recommend merge.

poorvith-mp:
Approved the design! Note that there is currently a merge conflict in `service/scripts/container_meta.py` against `main` around `named_value_is_ai(key, val)`. Once you rebase against latest `main` and resolve the conflict, this will be ready to land.

poorvith-mp (approved):
LGTM. Resolves merge conflict against main and exempts Claude Code agent/skill frontmatter configuration while keeping provenance scanning intact.

